PR #7, opened 2021-03-15
Auto seed generation and secret exportation

krtk6160:


charts/lnd/templates/NOTES.txt
```diff
@@ -21,9 +21,14 @@
...
 3. To get the macaroon, run:
-export MACAROON=$(kubectl exec -n {{ .Release.Namespace }} {{ .Release.Name }}-0 -- base64 /root/.lnd/data/chain/bitcoin/{{.Values.global.network}}/admin.macaroon | tr -d '\n\r')
+export MACAROON=$(kubectl exec -n {{ .Release.Namespace }} {{ include "lnd.fullname" . }}-0 -- base64 /root/.lnd/data/chain/bitcoin/{{.Values.global.network}}/admin.macaroon | tr -d '\n\r')
 4. To execute lncli against the pod, run the following commands:
-kubectl -n {{ .Release.Namespace }} port-forward {{ .Release.Name }}-0 10009
+kubectl -n {{ .Release.Namespace }} port-forward {{ include "lnd.fullname" . }}-0 10009
 lncli -n {{ .Release.Namespace }} help
+5. To retrieve the seed for the lnd wallet, run:
+kubectl -n {{ .Release.Namespace }} logs {{ include "lnd.fullname" . }}-wallet-create
+kubectl -n {{ .Release.Namespace }} delete pod {{ include "lnd.fullname" . }}-wallet-create
```
nicolasburtey, line 32: why delete the pod after?

charts/lnd/templates/cleanup-hook.yaml
```diff
@@ -0,0 +1,23 @@
+# Does not work currently
+# When the hook is disabled, the macaroon and tls configmaps are not deleted
+# When the hook is enabled, they get deleted right before the hook runs
```
nicolasburtey, line 3: why do you mean by they get deleted right before the hook runs? who's deleting it?

being deleted is what we want I guess.

charts/lnd/templates/export-secrets-configmap.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ printf "%s-export-secrets" (include "lnd.fullname" .) }}
+  labels:
+    {{- include "lnd.labels" . | nindent 4 }}
+data:
+  exportSecrets.sh: |
+    #!/bin/sh
+    kubectl wait --for=condition=ready --timeout=150s pod {{ include "lnd.fullname" . }}-0
```
nicolasburtey, line 10: could it take longer for the pod to start?
I guess if bitcoind is not ready we can still create the wallet?

charts/lnd/templates/export-secrets-configmap.yaml
```diff
@@ -0,0 +1,21 @@
...
+    kubectl cp {{ include "lnd.fullname" . }}-0:/root/.lnd/tls.cert ./tls.cert
+    kubectl create configmap {{ include "lnd.fullname" . }}-tls --from-file=./tls.cert --dry-run=client -o yaml | kubectl apply -f -
```
nicolasburtey, line 12: I think it should be a secret, not a configmap

charts/lnd/templates/lnd-pass-secret.yaml
```diff
@@ -4,5 +4,5 @@ metadata:
   name: {{ printf "%s-pass" (include "lnd.fullname" .) }}
   labels:
     {{- include "lnd.labels" . | nindent 4 }}
-stringData:
-  password: {{ default "randAlphaNum 24" .Values.walletPassword | quote }}
+data:
+  password: {{ template "walletPassword"  .  }}
```
nicolasburtey, line 8: it's already in base64?

charts/lnd/templates/NOTES.txt
krtk6160: To remove one trace of the seed, so that it's not very easy to retrieve it. Lncli does the same if I'm not wrong, meaning that once that screen goes away, you can't get it back.

charts/lnd/templates/_helpers.tpl
```diff
@@ -60,3 +60,19 @@ Create the name of the service account to use
...
 {{- end }}
+
+{{- define "walletPassword" -}}
+
+{{- $secret := (lookup "v1" "Secret" .Release.Namespace ( printf "%s-pass" (include "lnd.fullname" .))) -}}
+{{- if $secret -}}
+{{/*
+   Reusing current password since secret exists
+*/}}
+{{-  $secret.data.password -}}
```
nicolasburtey, line 71: this is the syntax to return a value?

charts/lnd/values.yaml
```diff
@@ -106,7 +106,7 @@ affinity:
             - bitcoind
         topologyKey: kubernetes.io/hostname
 
-walletPassword: 12345678
+walletPassword: ""
```
nicolasburtey, line 109: I guess an empty string would be defined as password is not set? (versus password is an empty string)? have you tested it?

nicolasburtey (commented):
good first pass. a couple of comments for clarification.

charts/lnd/templates/cleanup-hook.yaml
krtk6160: Consider these two scenarios:

- We don't have this cleanup hook enabled. We delete the lnd release. Now we have two configmaps (tls and macaroon) leftover.

- We have this cleanup hook enabled. We delete the lnd release. Now the hook fails by saying that the tls and macaroon configmaps do not exist. (They're getting deleted auto-magically right before this hook runs, I couldn't figure out by what process/job). Which means the `helm uninstall lnd` command gets stuck and ultimately times out.

charts/lnd/templates/export-secrets-configmap.yaml
krtk6160: > could it take longer for the pod to start?
> I guess if bitcoind is not ready we can still create the wallet?

Yep, wallet can be created before lnd tries to talk to bitcoind, but I'm not sure if lnd will become ready until it can talk to bitcoind. But if a user installs bitcoind, he will have to wait for a long time before installing lnd due to the time it takes for bitcoind to sync. So I guess we cannot really take the bitcoind delay into account here and increase the timeout because it could be days.

charts/lnd/templates/_helpers.tpl
krtk6160: It's the syntax to define a value, yes. `$secret` just refers to the result from the lookup.

charts/lnd/values.yaml
krtk6160: > I guess an empty string would be defined as password is not set? (versus password is an empty string)? have you tested it?

Yes I have tested it. Empty string returns false inside an `if`

charts/lnd/templates/lnd-pass-secret.yaml
krtk6160: The templates yes. I had missed adding the password from values.yaml so I have fixed that now.

charts/lnd/templates/NOTES.txt
nicolasburtey: you can from the db but there is no api for it.

charts/lnd/templates/cleanup-hook.yaml
nicolasburtey: > We delete the lnd release. Now the hook fails by saying that the tls and macaroon configmaps do not exist.

I'm probably missing something, but this would mean that we don't need the hook?

charts/lnd/templates/export-secrets-configmap.yaml
nicolasburtey: umbrel has a nice trick: [start lnd with neutrino, and switch when bitcoin is synced](https://github.com/getumbrel/umbrel/blob/79c011ea8ec9a97f3e68654f8d91dae048cc7b5b/scripts/configure#L209).

but I don't think we need this for what we do. we're targetting more large scale opeation, not a raspbery self usage solution.

charts/lnd/templates/cleanup-hook.yaml
krtk6160: > > We delete the lnd release. Now the hook fails by saying that the tls and macaroon configmaps do not exist.
> 
> I'm probably missing something, but this would mean that we don't need the hook?

That's the strange thing. Like I mentioned in the first scenario, if we don't have the hook, then the tls and macaroon do not get deleted. They are leftover despite all other lnd resources getting deleted.

So it seems like the hook only is responsible for this deletion, but for some reason it fails, and then goes into a crashloopbackoff, whereas you'd expect it to just get completed.

charts/lnd/templates/NOTES.txt
krtk6160: Yes. Basically the idea is to not add a way to retrieve the seed easily. Do you feel otherwise?

charts/lnd/templates/export-secrets-configmap.yaml
krtk6160: Fixed

charts/lnd/templates/NOTES.txt
nicolasburtey: I agree. should be written offline somewhere as a backup

the way to get the seed then would be (for us or an attacker):
- get the secret related to the passwod
- get access to the disk to get the encrypted boltdb (encrypted with the secret password above)

I believe one way to improve the security quite significantly would be to integrate [Hashicorp Vault](https://www.vaultproject.io/docs). 

The idea would be that the secret would be stored in memory only but not accessible by us (or at least not as easily as a Kubernetes Secret). But this would require time to think about it and document ourselves to Vault, so this is for another PR!

nicolasburtey (commented):
looks good to me. make sure to back things up before deploying this

nicolasburtey:
note: linting seems to fail

krtk6160:
> note: linting seems to fail

Yep, that is a bug with chart-testing and should be fixed in a few days. [Here's ](https://github.com/helm/chart-testing/pull/300) the discussion around it.

PR #9, opened 2021-03-19
Add thunderhub

krtk6160:


charts/lnd/Chart.yaml
```diff
@@ -16,8 +16,10 @@ engine: gotpl
 dependencies:
   - name: loop
     version: x.x.x
-  - name: rtl
+    condition: loop.enabled
```
nicolasburtey, line 19: didn't know we could do that. is that a new feature of helm?

charts/lnd/charts/thunderhub/templates/deployment.yaml
```diff
@@ -0,0 +1,82 @@
...
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
+          - name: ACCOUNT_CONFIG_PATH
+            value: "/home/thunderhub.yaml"
+          - name: LOG_LEVEL
+            value: 'debug'
+          - name: YML_ENV_1
```
nicolasburtey, line 38: we could use a better name

nicolasburtey (commented):
could we keep rtl and have both as an option?

rtl can be helpful if we ever want to have c-lightning as an additional lightning node option. also I've never tried thunderhub on my side so having both can be helpful for now to see how to best use thunderhub.

I know there was some auth/secret hurdle with rtl maybe this is why you have deleted it?

charts/lnd/Chart.yaml
krtk6160: Afaik it was there in helm v2 also, but the way to achieve it was a bit different. This was what I was referring to when we had the discussion over subchart enabling/disabling.

charts/lnd/charts/thunderhub/templates/deployment.yaml
krtk6160: Thunderhub requires us to use this env var name.

krtk6160:
> could we keep rtl and have both as an option?

> rtl can be helpful if we ever want to have c-lightning as an additional lightning node option. also I've never tried thunderhub on my side so having both can be helpful for now to see how to best use thunderhub.
> 

Okay.

> I know there was some auth/secret hurdle with rtl maybe this is why you have deleted it?

No, I deleted it because I thought we wanted to switch to thunderhub.

charts/lnd/charts/thunderhub/templates/deployment.yaml
nicolasburtey: ok

nicolasburtey (approved):

PR #15, opened 2021-04-30
Specter update

nicolasburtey:


charts/bitcoind/templates/networkpolicy-bitcoind.yaml
```diff
@@ -0,0 +1,13 @@
+kind: NetworkPolicy
```
krtk6160, line 1: Should be renamed to `bitcoind-networkpolicy.yaml` to conform to the conventions seen with bitnami charts.

PR #24, opened 2021-05-20
exporting pubkey as a secret

nicolasburtey:
we will need this for graphql to know which invoice are ours. using an env variable instead of dynamic fetch at runtime so that we know the pubkey even if a pod is down

### DIFF
--- a/charts/lnd/templates/export-secrets-configmap.yaml
+++ b/charts/lnd/templates/export-secrets-configmap.yaml
@@ -7,6 +7,10 @@ metadata:
 data:
   exportSecrets.sh: |
     #!/bin/sh
+
+    apk update
+    apk add expect curl ca-certificates openssl xxd
+
     kubectl wait --for=condition=ready --timeout=150s pod {{ include "lnd.fullname" . }}-0
     kubectl cp {{ include "lnd.fullname" . }}-0:/root/.lnd/tls.cert ./tls.cert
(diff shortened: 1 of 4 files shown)


charts/lnd/templates/export-secrets-configmap.yaml
```diff
@@ -23,3 +23,7 @@ data:
     export MACAROON=$(base64 ./macaroon/admin.macaroon | tr -d '\n\r')
     kubectl create secret generic {{ include "lnd.fullname" . }}-macaroon --from-literal=admin.macaroon=$MACAROON --dry-run=client -o yaml | kubectl apply -f -
     kubectl create secret generic {{ include "lnd.fullname" . }}-all-macaroons --from-file=./macaroon --dry-run=client -o yaml | kubectl apply -f -
+    
+    MACAROON_HEADER="Grpc-Metadata-macaroon: $(xxd -ps -u -c 1000 ./macaroon/admin.macaroon)" 
+    export PUBKEY=$(curl -s --cacert ./tls.cert --header "$MACAROON_HEADER"  https://{{ include "lnd.fullname" . }}:8080/v1/getinfo | jq .identity_pubkey)
```
krtk6160, line 28: Any reason for not using `kubectl exec` to get the pubkey? That command would look simpler and be a one-liner I think

krtk6160 (commented):
Couple of comments

krtk6160:
@nicolasburtey Can I take over this PR?

PR #28, opened 2021-05-23
Add lndmon and refactor lnd + subcharts

krtk6160:


charts/lnd/charts/lndmon/values.yaml
```diff
@@ -0,0 +1,86 @@
...
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: v0.2.0
+
+imagePullSecrets: []
+nameOverride: ""
+fullnameOverride: ""
+
+serviceAccount:
+  # Specifies whether a service account should be created
+  create: true
```
nicolasburtey, line 19: lndmon doesn't need a service account?

charts/lnd/charts/loop/Chart.yaml
```diff
@@ -15,9 +15,10 @@ type: application
...
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.3
+version: 0.1.4
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 0.11.1
+# It is recommended to use it with quotes.
```
nicolasburtey, line 23: why is that?

charts/lnd/charts/rtl/values.yaml
```diff
@@ -1,22 +1,69 @@
...
-  type: ClusterIP
+podAnnotations: {}
 
-podSecurityContext: 
-  fsGroup: 2000
+podSecurityContext: {}
+  # fsGroup: 2000
 
-securityContext:
+securityContext: {}
```
nicolasburtey, line 24: why deactivate non root access?

charts/lnd/templates/wallet-init-hook.yaml
```diff
@@ -14,7 +15,7 @@ spec:
     image: "boxboat/kubectl"
     imagePullPolicy: IfNotPresent
     command: ['/bin/sh']
-    args: ['-c','apk update; apk add expect; ./walletInit.exp {{.Values.global.network}} $LND_PASS {{ include "lnd.fullname" . }}-0;']
+    args: ['-c', 'apk update; apk add expect; kubectl wait --for=condition=ready --timeout=150s pod {{ include "lnd.fullname" . }}-0; ./walletInit.exp {{.Values.global.network}} $LND_PASS {{ include "lnd.fullname" . }}-0;']
```
nicolasburtey, line 18: the `.exp` is confusing to me. what is that supposed to mean?

charts/lnd/templates/wallet-init-hook.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   annotations:
     "helm.sh/hook": "post-install"
 spec:
+  # FIXME: should not run as root
   securityContext:
     runAsUser: 0
```
nicolasburtey, line 11: we could run it as non root with the new image I think?

nicolasburtey (commented):
solid PR. great to see this refactoring. couple of questions.

charts/lnd/charts/loop/Chart.yaml
krtk6160: Not sure. If you do `helm create xyz`, the Chart.yaml will be generated with this comment i.e. it's not a custom comment

charts/lnd/templates/wallet-init-hook.yaml
krtk6160: Any extension can be used, but tt's a convention for expect scripts to use `.exp` extension. More info [here](https://docs.oracle.com/cd/E35328_01/E35336/html/vmcli-script.html). 

It's a way of self documenting to make it clear that it's not a bash script.

charts/lnd/charts/rtl/values.yaml
krtk6160: Didn't realize it. Fixed.

charts/lnd/charts/lndmon/values.yaml
krtk6160: Fixed

charts/lnd/templates/wallet-init-hook.yaml
krtk6160: I tried it, didn't work. This hook actually isn't related to the lnd image at all, it just runs a bash script. The `expect` package needs to be installed, and for that we need to run this as a root user.

Issue #13, opened 2021-04-13
Fix the lnd init script


krtk6160:
Despite the wallet already being unlocked, the script keeps going into a loop (at least on devnet).

krtk6160:
Fixed by [this](https://github.com/GaloyMoney/charts/commit/0726be557891acce90b01f44e626135b3dd91ef5) commit

PR #32, opened 2021-06-11
Specify bitcoin network in lnd conf

bodymindarts:
When installing lnd via
```
$ helm install lnd galoy/lnd --set global.network=regtest
$ kubectl logs -f lnd-0
(...)
loadConfig: either --bitcoin.mainnet, or bitcoin.testnet, bitcoin.simnet, or bitcoin.regtest must be specified
(...)
```

shows that the config regarding which network connect to is missing.
After this is fixed it still doesn't start completely but I'm making tiny PRs addressing the issues step by step.

Related to https://github.com/GaloyMoney/charts/issues/13

### DIFF
--- a/charts/lnd/templates/configmap.yaml
+++ b/charts/lnd/templates/configmap.yaml
@@ -6,6 +6,7 @@ metadata:
     {{- include "lnd.labels" . | nindent 4 }}
 data:
   lnd.conf: |-
+    bitcoin.{{ .Values.global.network }}=true
   {{ printf "%s=%s" "bitcoind.rpcpass" (include "bitcoind.rpcpassword" . ) | indent 2 }}
   {{- range .Values.lndGeneralConfig }}
     {{ . }}


bodymindarts:
I'm just working on a Makefile that can bring up a dev environment via:
```
$ make dev
```

Or something like that.

These are the commands I'm running:
```
$ minikube start
$ helm repo add galoy https://galoymoney.github.io/charts/
$ helm install bitcoind galoy/bitcoind --values https://raw.githubusercontent.com/GaloyMoney/charts/main/charts/bitcoind/regtest-values.yaml
$ helm install lnd galoy/lnd --set global.network=regtest
$ source ./exportLocal.sh
$ ./initLocalTest.sh
```
But its not coming up flawlessly yet.

Long term I would suggest bringing the entire environment provisioning into a declarative format like terraform (it has a k8s and a helm provider) and re-use the same files for setting up dev / staging / prod for as much environment parity as possible.

chdwlch:
@bodymindarts some changes in https://github.com/GaloyMoney/galoy/pull/224 that you would find useful for successfully executing that sequence of commands

krtk6160:
> When installing lnd via
> 
> ```
> $ helm install lnd galoy/lnd --set global.network=regtest
> $ kubectl logs -f lnd-0
> (...)
> loadConfig: either --bitcoin.mainnet, or bitcoin.testnet, bitcoin.simnet, or bitcoin.regtest must be specified
> (...)
> ```
> 
> shows that the config regarding which network connect to is missing.

This is strange, we already have an entry for the `bitcoin.active=true` in our [values.yaml](https://github.com/GaloyMoney/charts/blob/main/charts/lnd/values.yaml#L119) file. So adding it again shouldn't be changing anything.

Could you post the output of `kubectl describe cm lnd`?

I have faced similar issues as you in the past, and almost everytime it was related to a permissions issue.

PR #34, opened 2021-06-24
Add tor container to lnd pod

krtk6160:


krtk6160:
> did you had a look at umbrel to see how they're going it?

Yep, but they don't use k8s. They use docker, and it [seems](https://github.com/getumbrel/umbrel/blob/master/docker-compose.yml) they have separate tor containers for each app that requires it.

PR #48, opened 2021-07-19
Initial concourse CI

bodymindarts:
Deployes a TestFlight environment and tears down again. Depends on https://github.com/GaloyMoney/charts/pull/45.

Closes https://github.com/GaloyMoney/charts/issues/47

### DIFF
--- a/charts/bitcoind/templates/secrets.yaml
+++ b/charts/bitcoind/templates/secrets.yaml
@@ -1,0 +1,11 @@
+{{ $rpcpassword := include "rpcpassword" . }}
+{{- if .Values.secrets.create }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: bitcoind-rpcpassword
+  labels:
+    {{- include "bitcoind.labels" . | nindent 4 }}
+stringData:
+  password: {{ $rpcpassword }}
+{{- end }}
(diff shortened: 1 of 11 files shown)


charts/bitcoind/templates/secrets.yaml
```diff
@@ -0,0 +1,11 @@
+{{ $rpcpassword := include "rpcpassword" . }}
```
krtk6160, line 1: You will have to change the `rpcpassword` block in `helpers.tpl` to remove the random pass generation. Otherwise this secret and the configmap will end up having a different value for the password.

bodymindarts: addressed here https://github.com/GaloyMoney/charts/pull/49

ci/repipe
```diff
@@ -0,0 +1,22 @@
...
+
+if [[ $(which ytt) == "" ]]; then
+  echo "You will need to install ytt to repipe. https://carvel.dev/ytt/"
+  exit 1
+fi
+
+target="${FLY_TARGET:-galoy}"
+team=dev
+
+TMPDIR=""¬
```
krtk6160, line 13: What does `¬` do here? Is that a key on your keyboard or did you have to get it from elsewhere? I don't have it on mine

bodymindarts: Good spot... its a copy paste mistake will get rid of it

Issue #44, opened 2021-07-19
Refactor handling of rpcpassword


bodymindarts:
Desired handling of rpc password:

can either be hard coded via values OR pre-provisioned via a secret (addressed [here](https://github.com/GaloyMoney/charts/pull/50)).

If its pre-provisioned the helm chart should not create the secret `secrets.create: false` (addressed [here](https://github.com/GaloyMoney/charts/pull/45)).

It should not be exposed via the `configmap` rather it must be added to `bitcoind.conf` via the init container step (addressed [here](https://github.com/GaloyMoney/charts/pull/49)).

PR #50, opened 2021-07-19
No rpcpassword generation

bodymindarts:
This enables determining the rpcpassword either via values.yml or via pre-provisioning a secret.

Password generation as part of the helm chart is not standard practice.

Potentially breaking change. Must specify:
```
secrets:
  create: false
```
In existing installations to not override the secret.

Build on https://github.com/GaloyMoney/charts/pull/49
Addresses https://github.com/GaloyMoney/charts/issues/44

### DIFF
--- a/charts/bitcoind/Chart.yaml
+++ b/charts/bitcoind/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.15
+version: 0.2.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/bitcoind/templates/_helpers.tpl
+++ b/charts/bitcoind/templates/_helpers.tpl
@@ -62,19 +62,3 @@ Create the name of the service account to use
 {{- end }}
 {{- end }}
 
-
-{{- define "rpcpassword" -}}
-
-{{- $secret := (lookup "v1" "Secret" .Release.Namespace "bitcoind-rpcpassword") -}}
-{{- if $secret -}}
-{{/*
-   Reusing current password since secret exists
-*/}}
-{{-  $secret.data.password | b64dec -}}
-{{- else -}}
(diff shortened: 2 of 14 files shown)


krtk6160:
> Password generation as part of the helm chart is not standard practice.

We have been trying to follow patterns from bitnami's helm charts (since they're the most widely used charts to my knowledge). They have password generation capabilities as part of lots of their charts. [Redis](https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/secret.yaml) and [mongodb](https://github.com/bitnami/charts/blob/master/bitnami/mongodb/templates/secrets.yaml), both of which we use also have this.

Is there a standard guide that is used within the helm community that says password generation is not the standard practice?

bodymindarts:
> > Password generation as part of the helm chart is not standard practice.
> 
> We have been trying to follow patterns from bitnami's helm charts (since they're the most widely used charts to my knowledge). They have password generation capabilities as part of lots of their charts. [Redis](https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/secret.yaml) and [mongodb](https://github.com/bitnami/charts/blob/master/bitnami/mongodb/templates/secrets.yaml), both of which we use also have this.
> 
> Is there a standard guide that is used within the helm community that says password generation is not the standard practice?

Interesting... I don't mind generating a password as a default. What I don't like is the `lookup` that dynamically talks to Kubernetes at deploy time.
That doesn't happen in the [bitnami charts](https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/_helpers.tpl#L204) either:
```
{{- define "redis.password" -}}
{{- if not (empty .Values.global.redis.password) }}
    {{- .Values.global.redis.password -}}
{{- else if not (empty .Values.auth.password) -}}
    {{- .Values.auth.password -}}
{{- else -}}
    {{- randAlphaNum 10 -}}
{{- end -}}
{{- end -}}
```

They also have the conditional to create / not create the secret set statically in the values file. Check the first line [here](https://github.com/bitnami/charts/blob/da613f5fa40810a915d615ed60b1edd8a829e313/bitnami/redis-cluster/templates/secret.yaml).

So the generation doesn't really have much effect. To pre-provision the secret or re-use the secret once its been deployed initially you have to change the `values.yml`.

PR #162, opened 2021-10-28
Dealer Helm Chart

sandipndev:
Adds the chart to spin up [GaloyMoney/dealer](https://github.com/GaloyMoney/dealer).

### DIFF
--- a/charts/dealer/Chart.yaml
+++ b/charts/dealer/Chart.yaml
@@ -1,0 +1,30 @@
+apiVersion: v2
+name: dealer
+description: A Helm chart for Kubernetes
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "0.1.0"
+
+dependencies:
+  - name: postgresql
+    repository: https://charts.bitnami.com/bitnami
+    version: 10.12.9
+    condition: postgresql.enabled
--- a/charts/dealer/Chart.yaml
+++ b/charts/dealer/Chart.yaml
@@ -21,7 +21,7 @@ version: 0.1.0
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "0.1.0"
+appVersion: 0.1.0
 
 dependencies:
   - name: postgresql
--- a/charts/dealer/templates/dealer-deployment.yaml
+++ b/charts/dealer/templates/dealer-deployment.yaml
@@ -30,23 +30,35 @@ spec:
         image: "{{ .Values.image.repository }}@{{ .Values.image.digest }}"
         env:
         - name: NETWORK
-          value: regtest
+          value: {{ .Values.network }}
         - name: LOG_LEVEL
-          value: debug
+          value: {{ .Values.log_level }}
         - name: ACTIVE_STRATEGY
-          value: OKEX_PERPETUAL_SWAP
+          value: {{ .Values.active_strategy }}
         - name: ACTIVE_WALLET
-          value: REMOTE_WALLET
+          value: {{ .Values.active_wallet }}
         - name: OKEX5_KEY
-          value: key
+          valueFrom:
+            secretKeyRef:
+              name: {{ .Values.okex5.secret_name }}
+              key: {{ .Values.okex5.secret_key.key }}
         - name: OKEX5_SECRET
-          value: secret
+          valueFrom:
+            secretKeyRef:
+              name: {{ .Values.okex5.secret_name }}
+              key: {{ .Values.okex5.secret_key.secret }}
         - name: OKEX5_PASSWORD
-          value: pwd
+          valueFrom:
+            secretKeyRef:
+              name: {{ .Values.okex5.secret_name }}
+              key: {{ .Values.okex5.secret_key.password }}
         - name: OKEX5_FUND_PASSWORD
-          value: none
+          valueFrom:
+            secretKeyRef:
+              name: {{ .Values.okex5.secret_name }}
+              key: {{ .Values.okex5.secret_key.fund_password }}
         - name: GRAPHQL_URI
-          value: "http://fake-galoyapi:4000/graphql"
+          value: {{ .Values.graphql_uri }}
         - name: DATABASE_URL
           valueFrom:
             secretKeyRef:
--- a/charts/dealer/templates/dealer-deployment.yaml
+++ b/charts/dealer/templates/dealer-deployment.yaml
@@ -21,44 +21,44 @@ spec:
         prometheus.io/path: /metrics
         prometheus.io/port: "3333"
         prometheus.io/scrape: "true"
-      {{- with .Values.annotations }}
+      {{- with .Values.dealer.annotations }}
         {{ toYaml . | nindent 8 }}
       {{- end }}
     spec:
       containers:
       - name: {{ .Chart.Name }}
-        image: "{{ .Values.image.repository }}@{{ .Values.image.digest }}"
+        image: "{{ .Values.dealer.image.repository }}@{{ .Values.dealer.image.digest }}"
         env:
         - name: NETWORK
-          value: {{ .Values.network }}
+          value: {{ .Values.dealer.network }}
         - name: LOG_LEVEL
-          value: {{ .Values.log_level }}
+          value: {{ .Values.dealer.logLevel }}
         - name: ACTIVE_STRATEGY
-          value: {{ .Values.active_strategy }}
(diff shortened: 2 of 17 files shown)


charts/dealer/Chart.yaml
```diff
@@ -0,0 +1,30 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
```
bodymindarts, line 24: this is obviously wrong - lets keep it at `0.1.0` until we start properly versioning the packaged system.

charts/dealer/templates/graphql-deployment.yaml
```diff
@@ -0,0 +1,30 @@
+{{ if (index .Values "fake-galoyapi").enabled }}
```
bodymindarts, line 1: I think we should keep with the idiomatic `camelCase` in our charts.

charts/dealer/templates/graphql-deployment.yaml
```diff
@@ -0,0 +1,30 @@
...
+    spec:
+      containers:
+      - name: graphql
+        image: "{{ (index .Values "fake-galoyapi").image.repository }}@{{ (index .Values "fake-galoyapi").image.digest }}"
+        ports:
+        - containerPort: 4000
+        env:
+        - name: DATABASE_URL
+          value: {{ (index .Values "fake-galoyapi").db_url }}
+{{- end }}
```
bodymindarts, line 30: Update your editor to add `\n` and EOF. This is not a valid unix file.

charts/dealer/templates/graphql-service.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if (index .Values "fake-galoyapi").enabled }}
```
bodymindarts, line 1: I think the file names should also have `fake` in them.

bodymindarts: Actually they should be called the same as the setting in values.yml. So this should be `fake-galoyapi-service.yml` etc.

charts/dealer/values.yaml
```diff
@@ -0,0 +1,33 @@
...
+  image:
+    repository: us.gcr.io/galoy-org/fake-galoyapi
+    digest: "sha256:684fccd8d64027a4e74dcd57a68af90a66cd31158f966c89878b90d756709b53" # Auto updated by Galoy Bot
+    git_ref: "" # Not used by helm, auto updated by Galoy Bot
+  service:
+    type: ClusterIP
+    port: 4000
+  db_url: "postgres://postgres:postgres@dealer-postgresql:5432/galoy"
+
+postgresql:
```
bodymindarts, line 29: I'm sure we can also pre-provision a secret here for user/password. That should be the default.

charts/dealer/values.yaml
```diff
@@ -0,0 +1,33 @@
...
+  enabled: true
+  postgresqlUser: postgres
+  postgresqlPassword: postgres
+  postgresqlDatabase: galoy
```
bodymindarts, line 33: Lets give it a name like `dealer` and reserve `galoy` for the main repo.

bodymindarts (changes requested):

charts/dealer/templates/graphql-service.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: {{ (index .Values "fake-galoyapi").name }}
+  labels:
+    app: {{ (index .Values "fake-galoyapi").name }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+spec:
+  type: {{ (index .Values "fake-galoyapi").service.type }}
+  ports:
+    - port: {{ (index .Values "fake-galoyapi").service.port }}
+      targetPort: http
```
krtk6160, line 14: Have you verified this service works?

charts/monitoring/Chart.lock
```diff
@@ -1,12 +0,0 @@
-dependencies:
```
bodymindarts, line 1: WTF

charts/dealer/templates/graphql-service.yaml
sandipndev: The ping goes through correctly.

charts/dealer/values.yaml
```diff
@@ -0,0 +1,32 @@
...
+  enabled: true
+  name: "fake-galoyapi"
+  image:
+    repository: us.gcr.io/galoy-org/fake-galoyapi
+    digest: "sha256:684fccd8d64027a4e74dcd57a68af90a66cd31158f966c89878b90d756709b53" # Auto updated by Galoy Bot
+    git_ref: "" # Not used by helm, auto updated by Galoy Bot
+  service:
+    type: ClusterIP
+    port: 4000
+  db_uri_secret_key_ref:
```
krtk6160, line 24: This should be `camelCase`

krtk6160 (changes requested):
A few comments

charts/dealer/templates/dealer-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+  selector:
+    matchLabels:
+      app: {{ template "dealer.fullname" . }}
+      release: {{ .Release.Name }}
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: {{ template "dealer.fullname" . }}
+        release: "{{ .Release.Name }}"
```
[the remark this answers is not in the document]
bodymindarts: I'd prefer to keep it. The more specific the selectors the less chance there is of accidental selecting more versions than necessary - its also how the concourse helm chart is done https://github.com/concourse/concourse-chart/blob/master/templates/web-deployment.yaml#L17-L19

charts/dealer/templates/fakegaloyapi-deployment.yaml
```diff
@@ -0,0 +1,33 @@
...
+  selector:
+    matchLabels:
+      app: {{ .Values.fakeGaloyApi.name }}
+      release: {{ .Release.Name }}
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: {{ .Values.fakeGaloyApi.name }}
+        release: "{{ .Release.Name }}"
```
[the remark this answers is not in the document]
bodymindarts: same as above

charts/dealer/templates/dealer-deployment.yaml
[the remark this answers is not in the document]
krtk6160: Could you explain how a scenario of "accidental selection" would pan out if we removed `release` here?

charts/dealer/Chart.yaml
```diff
@@ -0,0 +1,30 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "0.1.0"
```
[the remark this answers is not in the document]
sandipndev: Cool, changing the same for all the charts then.

krtk6160 (approved):

charts/dealer/templates/dealer-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+      {{- end }}
+    spec:
+      containers:
+      - name: {{ .Chart.Name }}
+        image: "{{ .Values.image.repository }}@{{ .Values.image.digest }}"
+        env:
+        - name: NETWORK
+          value: regtest
+        - name: LOG_LEVEL
+          value: debug
```
bodymindarts, line 35: Why debug by default?

charts/dealer/templates/dealer-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+        - name: ACTIVE_STRATEGY
+          value: OKEX_PERPETUAL_SWAP
+        - name: ACTIVE_WALLET
+          value: REMOTE_WALLET
+        - name: OKEX5_KEY
+          value: key
```
bodymindarts, line 41: What is this setting for?

charts/dealer/templates/dealer-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+        - name: OKEX5_SECRET
+          value: secret
```
bodymindarts, line 43: This can't be right or?

bodymindarts (changes requested):
The `env` section doesn't look complete. Or?

charts/dealer/templates/dealer-deployment.yaml
sandipndev: These were the defaults set in [.envrc](https://github.com/GaloyMoney/dealer/blob/main/.envrc)

sebastienverreault: the log level was "debug" for the local/dev setup or at least I set it so in the sample .envrc file
we can use "info" or whatever the default in the chart

charts/dealer/templates/dealer-deployment.yaml
sebastienverreault: ACTIVE_WALLET env. var. was so it could be switched between the remote (currently fake) wallet and the simulated version.
Simulated wouldn't make sense outside testing anyways so maybe default to remote in the code if env. var. is missing?

charts/dealer/templates/dealer-deployment.yaml
sebastienverreault: all OKEX5_* env. var. are dependent on the exchange account being used and are real secrets so the .envrc just had indication of those to get filled in

bodymindarts (approved):

PR #169, opened 2021-11-01
Remove admin panel as subchart

krtk6160:


charts/galoy/templates/admin-panel-deployment.yaml
```diff
@@ -0,0 +1,30 @@
+{{ if (index .Values "admin-panel").enabled }}
```
bodymindarts, line 1: what is `index` for here?

charts/galoy/templates/admin-panel-ingress.yaml
```diff
@@ -0,0 +1,35 @@
+{{- if (index .Values "admin-panel").ingress.enabled }}
+apiVersion: {{ include "common.capabilities.ingress.apiVersion" . }}
```
bodymindarts, line 2: Don't think this should be overridable. A change to the `apiVersion` would be a breaking change also for this template structure. Better to hard code it here to know what schema we need to apply.

charts/galoy/templates/admin-panel-ingress.yaml
```diff
@@ -0,0 +1,35 @@
...
+kind: Ingress
+metadata:
+  name: {{ include "admin-panel.fullname" . }}
+  labels:
+    app: {{ template "admin-panel.fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+  annotations:
+    {{- if (index .Values "admin-panel").ingress.certManager }}
+    kubernetes.io/tls-acme: "true"
```
bodymindarts, line 12: In other ingress files we have:
```
    kubernetes.io/ingress.class: nginx
    {{ if .Values.apiIngress.clusterIssuer }}
    cert-manager.io/cluster-issuer: {{ .Values.apiIngress.clusterIssuer }}
    {{ end }}
```
Is this equivilent?

charts/galoy/values.yaml
```diff
@@ -97,24 +97,22 @@ redis:
   metrics:
     enabled: true
 admin-panel:
-  enabled: false
+  enabled: true
+  nameOverride:
```
bodymindarts, line 101: Why is this set here?

charts/galoy/values.yaml
```diff
@@ -97,24 +97,22 @@ redis:
...
+  image:
+    repository: us.gcr.io/galoy-org/admin-panel
+    digest: "sha256:bd4f821f3c7e2cff57361f2f52ef85855585c42c997750a37ef7df970cfb3377"
+    git_ref: "3666aa3" # Not used by helm
+  labels: {}
```
bodymindarts, line 106: Why is this set?

charts/galoy/values.yaml
```diff
@@ -97,24 +97,22 @@ redis:
...
-    # there seems to be inconsistency on sysctls on bitnami
-    # sometimes it's a map, sometimes an array
-    sysctls: []
+    port: 80
   ingress:
     enabled: false
-  podLabels:
-    app: admin-panel
-  command: ['/bin/sh', '-c', 'echo "window.env={\"GRAPHQL_URI\":\"https://graphql.{{.Values.global.network}}.galoy.io/graphql\"}" > /app/env.js && /opt/bitnami/scripts/nginx/entrypoint.sh /opt/bitnami/scripts/nginx/run.sh']
+    certManager: false
```
bodymindarts, line 112: This doesn't make sense to me - why we setting certManager = false but adding the annotations? I think the boundaries are wrong. The annotations should be in the ingress definition or?

bodymindarts (changes requested):

charts/galoy/templates/admin-panel-deployment.yaml
krtk6160: `{{if .Values.admin-panel.enabled}}` is invalid helm syntax due to the hyphen

bodymindarts: Shouldn't we be calling it `adminPanel`? I think `camelCase` is idiomatic for helm charts right?

charts/galoy/templates/admin-panel-ingress.yaml
krtk6160: Fixed

charts/galoy/templates/admin-panel-ingress.yaml
krtk6160: The tls-acme annotation seems to be deprecated, so removing it and moving to the pattern being used in apiIngress

PR #193, opened 2021-11-05
Dev charts for JupyterHub

jireva:
Currently does not have access to any data.

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -34,6 +34,6 @@ To test the jupyter lab:
 kubectl -n galoy-dev-data port-forward service/proxy-public 8080:80
 ```
 
-Then open the browser to: http://127.0.0.1:8000/
+Then open the browser to: http://127.0.0.1:8080/
 
 Currently incomplete functionality - but depending on what you want to hack on it'll work
--- a/dev/README.md
+++ b/dev/README.md
@@ -33,7 +33,13 @@ Currently incomplete functionality - but depending on what you want to hack on i
 
 ## JupyterLab
 
-To test the jupyter lab:
+Add the JupyterHub Helm chart repository:
+```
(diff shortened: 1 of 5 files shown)


dev/data/config.yaml
```diff
@@ -0,0 +1,10 @@
+singleuser:
+  image:
+    # You should replace the "latest" tag with a fixed version from:
```
bodymindarts, line 3: Is this comment required?

dev/README.md
```diff
@@ -29,4 +29,11 @@ $ curl 'localhost:8080/graphql' -H 'Content-Type: application/json' -H 'Accept:
 {"data":{"userLogin":{"authToken":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1aWQiOiI2MTc2YmQ2NmQ0MmFkYWIzNjM2MmEyY2QiLCJuZXR3b3JrIjoibWFpbm5ldCIsImlhdCI6MTYzNTE3MTY4Nn0.n-p5sA9meAmZrVOdngYr216jG3LKOFsFdJmVw6XND3A"}}}
 ```
 
+To test the jupyter lab:
```
bodymindarts, line 32: Can you move this to its own subheading lower in the README.

bodymindarts (commented):
As we are not actually creating a chart for this it won't be necessary to add a TestFlight step. We can add it to staging directly in the deployments repo once we have the `addons` lifecycle.

dev/data/config.yaml
jireva: Removed it.

dev/README.md
jireva: Done.

bodymindarts (dismissed):

PR #201, opened 2021-11-06
Update galoy-values.yml

nicolasburtey:


bodymindarts (approved):

bodymindarts:
Why disable? Its good if its _enabled_ by default in dev or?

PR #234, opened 2021-11-14
Added proper smoketest for dealer along with it's service

sandipndev:
Adds:
1. A dealer internal service to contact the dealer deployment and reach the metrics endpoint
2. The proper smoketest that does 15 attempts on the metrics endpoint to fetch the exported values


charts/dealer/templates/dealer-service.yml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: {{ template "dealer.fullname" . }}
+  labels:
+    app: {{ template "dealer.fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+spec:
+  type: ClusterIP
+  ports:
+    - port: 80
```
krtk6160, line 12: Ports values should be taken from values.yaml

krtk6160 (changes requested):

krtk6160 (approved):

PR #286, opened 2021-11-30
Fix fake galoy api svc

krtk6160:


sandipndev (approved):
Tested, working, thanks for the fix!

PR #295, opened 2021-12-01
Fix pg password key name

krtk6160:


bodymindarts (approved):

sandipndev (commented):
Isn't this `admin_password`?
https://github.com/GaloyMoney/galoy-org-infra/blob/main/pipeline-creds/provision/postgres.tf#L8

sandipndev (approved):
Okay, this is correct.

PR #301, opened 2021-12-02
Adding faster scrape config for dealer

sandipndev:


charts/dealer/templates/dealer-deployment.yaml
```diff
@@ -20,7 +20,7 @@ spec:
       annotations:
         prometheus.io/path: {{ .Values.dealer.exporter.path }}
         prometheus.io/port: "{{ .Values.dealer.exporter.port }}"
-        prometheus.io/scrape: "true"
+        galoy.io/dealer: "true"
```
krtk6160, line 23: Is this a typo? Why has the annotation been changed?

charts/monitoring/values.yaml
```diff
@@ -196,6 +196,37 @@ prometheus:
               action: replace
               target_label: namespace
 
+        - job_name: 'lnd'
```
krtk6160, line 199: There's a typo here

krtk6160 (changes requested):

charts/monitoring/values.yaml
sandipndev: Fixed, thanks for noticing!

krtk6160 (approved):

PR #311, opened 2021-12-06
Update lnd image

krtk6160:


charts/lnd/templates/statefulset.yaml
```diff
@@ -161,7 +162,7 @@ spec:
               command:
               - sh
               - -c
-              - "lncli -n {{ $.Values.global.network}} state | grep -q RPC_ACTIVE"
+              - "lncli --lnddir=/data/.lnd -n {{ $.Values.global.network}} state | grep -q RPC_ACTIVE"
```
bodymindarts, line 165: Don't like the grep here... also this should be a special case, by default we should not be passing this its only required for migration.

Perhaps we could have:
values.yml
```
overrideDataDirPath: /data/.lnd/mainnet
```

when empty we don't pass that flag - when it exists (for the environments needing migration) we add it.

Ideally we need to find a way to make the change the path on the volume to be compatible with the new image. This is not code we want to maintain.

PR #328, opened 2021-12-09
Renamed staging worker to galoy-staging

sandipndev:


ci/pipeline-fragments.lib.yml
```diff
@@ -55,14 +55,14 @@ plan:
     run:
       path: pipeline-tasks/ci/tasks/prepare-testflight.sh
 - put: #@ "tf-" + chart + "-testflight"
-  tags: ["staging"]
```
krtk6160, line 58: We should be templating this var so that in case we want to change it again, we won't have to change it in numerous places.

sandipndev: Agreed, doing that.

bodymindarts (approved):

PR #333, opened 2021-12-13
Add galoy pay env vars

krtk6160:


bodymindarts (dismissed):

bodymindarts:
Shouldn't there be defaults set in `values.yml` for discoverability?

krtk6160:
> Shouldn't there be defaults set in `values.yml` for discoverability?

Yep. I'll add that and also refactor this a bit once the [refactor PR](https://github.com/GaloyMoney/galoy-pay/pull/142) is merged

PR #519, opened 2022-01-26
feat: added kubemonkey labels for lnd

sandipndev:
Could help in regular database compaction


charts/lnd/values.yaml
```diff
@@ -9,6 +9,9 @@ image:
   tag: v0.14.1-beta
   pullPolicy: IfNotPresent
 
+kubemonkey:
+  enabled: true
```
bodymindarts, line 13: default should be false. We only want on our environments

bodymindarts (approved):

PR #539, opened 2022-01-31
feat: add web wallet chart

krtk6160:
It's unclear at the moment what the ideal prod values will be for web-wallet, so for now I've just kept the values that make it work locally. Once deployed, I will change the values as required to match prod values.

### DIFF
--- a/charts/web-wallet/templates/deployment.yaml
+++ b/charts/web-wallet/templates/deployment.yaml
@@ -30,7 +30,10 @@ spec:
         - name: NODE_ENV
           value: production
         - name: SESSION_KEYS
-          value: {{ .Values.sessionKeys }}
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "webWallet.fullname" . }}
+              key: session-keys
         - name: HOST
           value: {{ .Values.host }}
         - name: PORT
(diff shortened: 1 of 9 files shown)


charts/web-wallet/templates/deployment.yaml
```diff
@@ -0,0 +1,45 @@
...
+      containers:
+      - name: web-wallet
+        image: "{{ .Values.image.repository }}@{{ .Values.image.digest }}"
+        ports:
+        - containerPort: {{ .Values.service.port }}
+        env:
+        - name: NODE_ENV
+          value: production
+        - name: SESSION_KEYS
+          value: {{ .Values.sessionKeys }}
```
bodymindarts, line 33: I believe these are sensitive and should be injected via a secret

charts/web-wallet/values.yaml
```diff
@@ -0,0 +1,23 @@
...
+  git_ref: "8189266" # Not used by helm
+
+ingress:
+  enabled: false
+
+service:
+  port: 3000
+  type: ClusterIP
+
+authEndpoint: https://example.com/api/login
```
bodymindarts, line 13: This is fine for now - would be cool if defaults work out of the box locally

bodymindarts (approved):

bodymindarts (approved):

PR #585, opened 2022-02-08
feat: add network policies

krtk6160:


charts/galoy/templates/cronjob.yaml
```diff
@@ -5,9 +5,15 @@ metadata:
   name: {{ .name }}
   labels:
     app.kubernetes.io/managed-by: Helm
+    app: {{ .name }}
 spec:
   schedule: {{ .schedule }}
   jobTemplate:
+    metadata:
+      {{- with .podLabels }}
```
bodymindarts, line 13: This should not be exposed. The network policy related labels should be hard coded.

bodymindarts (approved):

charts/galoy/templates/cronjob.yaml
krtk6160: Fixed

PR #589, opened 2022-02-09
chore: add lnd sidecar image build job

krtk6160:


bodymindarts:
lets refactor and put this (+ chain-dl-image) under `<root>/images/<image>/Dockerfile`

ci/lnd-sidecar-image/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM alpine:3.15.0
```
nicolasburtey, line 1: is it possible to have a distroless image?

moving every Docker image to a distroless container would also improve the security

krtk6160: I tried to have the distroless be the base image, but ran into issues with shared libraries that were missing from the distroless image but required by our dependencies.

PR #605, opened 2022-02-14
feat: add galoy auth chart

krtk6160:


charts/galoy-auth/charts/auth-backend/values.yaml
```diff
@@ -0,0 +1,23 @@
...
+ingress:
+  enabled: false
+service:
+  port: 80
+  type: ClusterIP
+authEndpoint: api/login
+graphqlUri: http://localhost:4002/graphql
+graphqlSubscriptionUri: wss://localhost:4002/graphql
+supportEmail: support@galoy.io
+host: "0.0.0.0"
```
bodymindarts, line 14: why is this exposed?

charts/galoy-auth/charts/auth-backend/values.yaml
```diff
@@ -0,0 +1,23 @@
+image:
+  repository: us.gcr.io/galoy-org/web-wallet
+  digest: "sha256:ab4f308adba7241ea4ab71bc520b7e4fa404ba86fc65f77bb9d37d0ad19b63fc"
+  git_ref: "c9df056" # Not used by helm
+ingress:
+  enabled: false
+service:
+  port: 80
+  type: ClusterIP
+authEndpoint: api/login
```
bodymindarts, line 10: I think this should be the whole url not just a path

bodymindarts (approved):

charts/galoy-auth/charts/auth-backend/values.yaml
krtk6160: removed

charts/galoy-auth/charts/auth-backend/values.yaml
krtk6160: Not sure if this is being used, or will be used in auth backend. Will fix in a later PR if it's used.

PR #635, opened 2022-02-21
feat: added price server host and port env to galoy pod

sandipndev:


charts/galoy/templates/deployment.yaml
```diff
@@ -159,6 +159,17 @@ spec:
             - name: REDIS_2_DNS
               value: "galoy-redis-node-2.galoy-redis-headless"
 
+            - name: PRICE_SERVER_PORT
+              valueFrom:
+                secretKeyRef:
+                  name: price-secret
+                  key: port
+            - name: PRICE_SERVER_HOST
```
bodymindarts, line 167: these are not sensitive - read from values or hard code - not from secret

PR #693, opened 2022-03-04
feat: add web wallet mobile to testflight

krtk6160:


ci/testflight/web-wallet/main.tf
```diff
@@ -45,6 +46,18 @@ resource "helm_release" "web_wallet" {
   namespace  = kubernetes_namespace.testflight.metadata[0].name
 }
 
+resource "helm_release" "web_wallet" {
+  name       = "web-wallet-mobile"
+  chart      = "${path.module}/chart"
+  repository = "https://galoymoney.github.io/charts/"
+  namespace  = kubernetes_namespace.testflight.metadata[0].name
+
+  set {
```
bodymindarts, line 55: I'd prefer not to use `set` but consistently use a `*-values.yml` file even if its just for 1 setting. Its less confusing when finding where to change things.

krtk6160: Fixed

bodymindarts (approved):

PR #729, opened 2022-03-10
feat: add price history resources

krtk6160:


bodymindarts (approved):

charts/galoy/charts/price/templates/history-cronjob.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: {{ include "priceHistory.fullname" . }}-cronjob
+  labels:
+    app.kubernetes.io/managed-by: Helm
+spec:
+  schedule: "0 * * * *"
```
dolcalmi, line 8: some times exchanges data update take some secs/min so I think is better to run it in min 1 or 2

dolcalmi (changes requested):
looks good, just minor change in cron

charts/galoy/charts/price/templates/history-cronjob.yaml
krtk6160: Fixed

dolcalmi (approved):

PR #746, opened 2022-03-17
refactor: introduce top level webWallet key

bodymindarts:
Let us start having a top level `<application>:` key to group the app specific settings instead of putting so many settings loosely on top level.

### DIFF
--- a/charts/web-wallet/values.yaml
+++ b/charts/web-wallet/values.yaml
@@ -2,7 +2,8 @@ image:
   repository: us.gcr.io/galoy-org/web-wallet
   digest: "sha256:438220a9a01f0ce2ece49ae905bb073a5da43eef36cdbbeb5dfde6e9019d30c7"
   git_ref: "ec5dc40" # Not used by helm
-mobileLayoutImage:
+mobileLayout:
+  enabled: false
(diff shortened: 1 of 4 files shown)


charts/web-wallet/values.yaml
```diff
@@ -5,18 +5,27 @@ image:
 mobileLayoutImage:
   repository: us.gcr.io/galoy-org/web-wallet-mobile-layout
   digest: "sha256:40e4edbd98cb3d58bda687265fbf0230c6c80745e69778db15218c287ee0e18f"
-useMobileLayout: false
```
krtk6160, line 8: Shouldn't this be kept at top level? This is only used by helm

krtk6160 (commented):
One question

charts/web-wallet/values.yaml
bodymindarts: I contemplated that. The thing is I don't like it as a lone top level config (everything else is nested) and I don't know a better key to scope it under.

krtk6160: Perhaps we can have a `mobileLayout` top level field:

```
mobileLayout:
  enabled: boolean
  image:
    repo: string
    digest: string  
```

krtk6160 (approved):

PR #770, opened 2022-03-30
chore(deps): update helm chart dependencies

galoybot:
This PR updates Helm Chart Dependencies.

PR #772, opened 2022-03-31
ci: automate version bumping for helm deps

sandipndev:
Adds a CI task that checks for a newer dependency version for all of our Helm Charts and creates a PR to update them.
Example PR: https://github.com/GaloyMoney/charts/pull/770


sandipndev:
We need to make this in such a way that we check each release of helm charts separately, will be done in a separate PR.

PR #886, opened 2022-04-28
feat: array input for lnd values in ride the lightning config

zoop-btc:
This PR changes the configuration map to iterate over an array of lnd values. I have also added the possibility to enable dark mode and themes.
I've also set the channelBackupPath to a place where the rtl process has write access - else you get error messages with each channel open or close.

**!This changes the values structure!**
Here is an example for two lnds in the new format:
```
lnds:
  - url: lnd1
    secretName: lnd1-credentials
    dark: true
    themeColor: YELLOW
  - url: lnd2
    secretName: lnd2-credentials
    dark: false
    themeColor: BLUE
```
The possible values for themeColor are:  BLUE GREEN INDIGO PINK PURPLE TEAL YELLOW

UPDATE: Updated to match new naming convention.

### DIFF
--- a/charts/rtl/values.yaml
+++ b/charts/rtl/values.yaml
@@ -32,9 +32,11 @@ service:
   port: 3000
   type: ClusterIP
 
-# FIXME: Intake an array of lnds and secret names
-lnd1Url: lnd1
-lnd1SecretName: lnd1-credentials
+lnds:
+  - url: lnd1
+    secretName: lnd1-credentials
+    dark: true
+    themeColor: YELLOW
 
 resources: {}
 
--- a/charts/rtl/templates/configmap_and_secret.yaml
+++ b/charts/rtl/templates/configmap_and_secret.yaml
@@ -18,40 +18,28 @@ data:
         "logoutRedirectLink": "/login"
       },
       "nodes": [
+        {{- $c := 0 | int }}
+        {{- range .Values.lnds }}
+        {{- $c = add1 $c }}
         {
-          "index": 1,
-          "lnNode": "Node 1",
+          "index": {{ $c }},
+          "lnNode": "Node {{ $c }}",
           "lnImplementation": "LND",
           "Authentication": {
-            "macaroonPath": "/lnd1/rpc",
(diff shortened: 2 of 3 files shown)


bodymindarts:
It looks good except for the ingress. Exposing RTL on the internet is pretty bad from a security POV and I don't think its something we should support OOTB.

zoop-btc:
@bodymindarts I removed the ingress and cleaned up the pull request.

zoop-btc:
> @zoop-btc this would be nice to squash those commits

yes sorry, I wanted to bring this branch up to speed and it somehow added all the commits as new ones

fixed now

charts/rtl/values.yaml
```diff
@@ -32,9 +32,11 @@ service:
   port: 3000
   type: ClusterIP
 
-# FIXME: Intake an array of lnds and secret names
-lnd1Url: lnd1
-lnd1SecretName: lnd1-credentials
+lnd:
+  - url: lnd1
+    secretname: lnd1-credentials
```
krtk6160, line 37:
```suggestion
    secretName: lnd1-credentials
```

charts/rtl/values.yaml
```diff
@@ -32,9 +32,11 @@ service:
   port: 3000
   type: ClusterIP
 
-# FIXME: Intake an array of lnds and secret names
-lnd1Url: lnd1
-lnd1SecretName: lnd1-credentials
+lnd:
```
krtk6160, line 35:
```suggestion
lnds:
```

krtk6160 (changes requested):
@zoop-btc I have tested this locally and it works perfectly! Thank you for the PR. Just a couple of nits around naming, but otherwise LGTM

charts/rtl/values.yaml
zoop-btc: Renamed according to suggestion.

zoop-btc:
@krtk6160 done, adjusted the values, configmap and deployment accordingly

krtk6160 (approved):

PR #978, opened 2022-05-17
feat: add nextcloud compatibility and firebase optionality

zoop-btc:
I would like to propose these changes so it'll be possible to use Nextcloud as a backup mechanism.
This also introduces a new structure to the values. It makes sense to consolidate them as backup since they're both used for the mongodb as well as the scb backup. This is what I came up with but please feel free to give input!
```
backup:
  googlecloud:
    enabled: true
    bucketName: galoy-staging-backups
    secretName: gcs-sa-key
  dropbox:
    enabled: true
    secretName: dropbox-access-token
  nextcloud:
    enabled: false
    secretName: nextcloud-credentials
```

If a user wants to deploy the charts right now without having access to dropbox/google cloud/firebase then it will fail since the secrets referenced don't exist. The only solution I saw was to encapsulate them in if statements and use _firebaseaccountneeded_ as well as the _enabled_ values above. I made it so it'll still reflect the defaults by disabling nextcloud. However disabling firebase will turn off the [daily notifications](https://github.com/GaloyMoney/galoy/blob/main/src/servers/daily-balance-notification.ts), which anybody deciding to forego google needs to be made aware of. Maybe I will tackle that later on.

**Are there any other issues which not using google at all could create?** (I have not noticed any in my test environment.)

This pull request coincides with one [for the galoy repo ](https://github.com/GaloyMoney/galoy/pull/1316).

The secret for nexcloud should look like this and needs to be manually applied like the others:
```
apiVersion: v1
kind: Secret
metadata:
  name: nextcloud-credentials
type: Opaque
data:
  host: <base64encodedfullurl>
  user: <base64encodeduser>
  password: <base64encodedpassword>
```

The host should point to the full url of the nextcloud webdav target like this:
`https://myserver.com/remote.php/dav/files/myuser/backupfolder`

thank you kindly for the consideration
ps: sorry adding so many checks to the mongodb backup script, but better safe than sorry

### DIFF
--- a/charts/galoy/templates/mongo-backup-configmap.yaml
+++ b/charts/galoy/templates/mongo-backup-configmap.yaml
@@ -11,16 +11,37 @@ data:
     exit 0
     {{- end }}
     echo "backup script starts"
-    export BUCKET_NAME="{{ .Values.mongobackup.bucketName }}"
+    if [ -z "${DROPBOX_ACCESS_TOKEN}" ] && [ -z "${GOOGLE_APPLICATION_CREDENTIALS}" ] && ( [ -z "${NEXTCLOUD_USER}" ] || [ -z "${NEXTCLOUD_PASSWORD}" ] || [ -z "${NEXTCLOUD_URL}" ] ); then
+    echo "No environment variables for any backup destination set, aborting backup."
+    exit 1
+    fi
+    export BUCKET_NAME="{{ .Values.backup.googlecloud.bucketName }}"
     set -e
+    if [ -z "${MONGODB_USER}" ]; then echo "MONGODB_USER not set, aborting backup" && exit 1; fi
+    if [ -z "${MONGODB_ADDRESS}" ]; then echo "MONGODB_ADDRESS not set, aborting backup" && exit 1; fi
+    if [ -z "${MONGODB_PORT}" ]; then echo "MONGODB_PORT not set, aborting backup" && exit 1; fi
     LOG_TIME=$(date +%s)
     BACKUP_NAME="$NETWORK-$LOG_TIME.gz"
-    echo "Activating service account"
-    gcloud auth activate-service-account --key-file=$GOOGLE_APPLICATION_CREDENTIALS
     echo "Backing up mongodb"
     mongodump --host=$MONGODB_ADDRESS --port=$MONGODB_PORT --username=$MONGODB_USER --password=$MONGODB_PASSWORD --gzip --archive=$BACKUP_NAME -d=$MONGODB_DB --readPreference=secondary
+    if [ -z "${DROPBOX_ACCESS_TOKEN}" ]; then
+    echo "$DROPBOX_ACCESS_TOKEN not set, skipping dropbox upload."
+    else
     echo "Uploading backup $BACKUP_NAME to dropbox"
     curl -X POST https://content.dropboxapi.com/2/files/upload --http1.1 --header "Authorization: Bearer $DROPBOX_ACCESS_TOKEN" --header "Dropbox-API-Arg: {\"path\": \"/mongo/$BACKUP_NAME\"}" --header "Content-Type: application/octet-stream" --data-binary $BACKUP_NAME
+    fi
+    if [ -z "${GOOGLE_APPLICATION_CREDENTIALS}" ] || [ -z "${BUCKET_NAME}" ]; then
+    echo "GOOGLE_APPLICATION_CREDENTIALS or BUCKET_NAME not set, skipping google cloud upload."
+    else
+    echo "Activating service account"
+    gcloud auth activate-service-account --key-file=$GOOGLE_APPLICATION_CREDENTIALS
     echo "Uploading backup $BACKUP_NAME to gcs"
     gsutil cp $BACKUP_NAME gs://$BUCKET_NAME/mongodb/$BACKUP_NAME 2>&1
+    fi
+    if [ -z "${NEXTCLOUD_USER}" ] || [ -z "${NEXTCLOUD_PASSWORD}" ] || [ -z "${NEXTCLOUD_URL}" ]; then
+    echo "NEXTCLOUD_USER, NEXTCLOUD_PASSWORD or NEXTCLOUD_URL not set, skipping nextcloud upload."
+    else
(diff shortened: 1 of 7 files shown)


nicolasburtey:
thanks for your contribution @zoop-btc !

I didn't dive into the code, but making firebase a non mandatory dependency is very useful

charts/galoy/templates/mongo-backup-configmap.yaml
```diff
@@ -11,16 +11,40 @@ data:
...
+    exit 1
+    fi
+    export BUCKET_NAME="{{ .Values.backup.googlecloud.bucketName }}"
     set -e
+    MONGO_TOOLS_VERSION="{{ .Values.backup.mongoToolsVersion }}"
+    if [ -z "${MONGO_TOOLS_VERSION}" ]; then echo "MONGO_TOOLS_VERSION not set, aborting backup" && exit 1; fi
+    if [ -z "${MONGODB_USER}" ]; then echo "MONGODB_USER not set, aborting backup" && exit 1; fi
+    if [ -z "${MONGODB_ADDRESS}" ]; then echo "MONGODB_ADDRESS not set, aborting backup" && exit 1; fi
+    if [ -z "${MONGODB_PORT}" ]; then echo "MONGODB_PORT not set, aborting backup" && exit 1; fi
+    apk add --update --no-cache mongodb-tools=${MONGO_TOOLS_VERSION} --repository=http://dl-cdn.alpinelinux.org/alpine/edge/community
```
krtk6160, line 25:
```suggestion

```

charts/galoy/templates/mongo-backup-configmap.yaml
```diff
@@ -11,16 +11,40 @@ data:
...
     echo "backup script starts"
-    export BUCKET_NAME="{{ .Values.mongobackup.bucketName }}"
+    if [ -z "${DROPBOX_ACCESS_TOKEN}" ] && [ -z "${GOOGLE_APPLICATION_CREDENTIALS}" ] && ( [ -z "${NEXTCLOUD_USER}" ] || [ -z "${NEXTCLOUD_PASSWORD}" ] || [ -z "${NEXTCLOUD_URL}" ] ); then
+    echo "No environment variables for any backup destination set, aborting backup."
+    exit 1
+    fi
+    export BUCKET_NAME="{{ .Values.backup.googlecloud.bucketName }}"
     set -e
+    MONGO_TOOLS_VERSION="{{ .Values.backup.mongoToolsVersion }}"
+    if [ -z "${MONGO_TOOLS_VERSION}" ]; then echo "MONGO_TOOLS_VERSION not set, aborting backup" && exit 1; fi
```
krtk6160, line 21:
```suggestion

```

krtk6160 (changes requested):
The code involving mongo backup tools can be removed as we have now shifted to using a dedicated docker image for the mongo backup cronjob which has the dependency preinstalled

zoop-btc:
Thank you @krtk6160 this looks even better now since the mongodb-tools value can be completely removed :)
I've also cleaned up the pull request, sorry new to github.

zoop-btc:
I have moved the trigger outside of the deployment loop @bodymindarts  as per your suggestion.
This means another change to the values:
https://github.com/GaloyMoney/charts/blob/d2b61bf1212e5668448baf18dab492e33de452b2/charts/galoy/values.yaml#L46-L56

This format would be followed by the other galoy deployments. I can do it for the others as well in a separate PR.
I would also like to make the resources configurable there instead of having them fixed in the template.

krtk6160:
> I have moved the trigger outside of the deployment loop @bodymindarts as per your suggestion. This means another change to the values:
> 
> https://github.com/GaloyMoney/charts/blob/d2b61bf1212e5668448baf18dab492e33de452b2/charts/galoy/values.yaml#L46-L56
> 
> This format would be followed by the other galoy deployments. I can do it for the others as well in a separate PR. I would also like to make the resources configurable there instead of having them [fixed in the template](https://github.com/GaloyMoney/charts/blob/c4cf00c4dc72f8342ff765467f0ca8f275fbb9b8/charts/galoy/templates/deployment.yaml#L65).

@sandipndev iirc you too were refactoring the galoy chart. Does that have any overlap/conflict with the changes here? Might make sense to sync with @zoop-btc to avoid duplicated efforts or conflicts

sandipndev:
@krtk6160 nope, I was doing the secrets so it's not in the scope of this PR at all.

zoop-btc:
updated branch - no breaking changes found - ready for review

charts/galoy/values.yaml
```diff
@@ -74,12 +76,21 @@ cron:
...
-  mongoToolsVersion: 4.2.14-r6
   image:
     repository: us.gcr.io/galoy-org/mongo-backup
     digest: "sha256:4cfc197ea082ef3dae2d4321a9dfa31ef185f43d578a37a224095b75de656e6a"
     git_ref: 872b70e
+backup:
+  googlecloud:
+    enabled: true
+    bucketName: galoy-staging-backups
+    secretname: gcs-sa-key
```
krtk6160, line 87: minor nit: should use camelcase here -> `secretName` to be in line with other key names

zoop-btc: Thank you, is ammended now.

krtk6160:
@zoop-btc There's a significant refactor of the galoy chart that @sandipndev is doing in #1193. It has some amount of overlap with this PR, so those parts will no longer be necessary once we rebase this PR following the merge of #1193.

P.S. No need to rebase now, #1193 is still WIP.

zoop-btc:
@krtk6160 Understood, I'll wait for it to be merged and then adjust this PR accordingly. Thank you for the heads-up.
I'm putting it back to draft mode.

zoop-btc:
closing this, I will split it into two pull requests, one for nextcloud and the other to make firebase optional - I think factoring trigger out of the deployment was too big of a scope

sandipndev:
@zoop-btc the newer refactored charts have been merged. It now allows you to add in another `nextcloud` and `firebase` fields to https://github.com/GaloyMoney/charts/blob/main/charts/galoy/values.yaml#L184 and the overall work would be much cleaner.

PR #1012, opened 2022-05-24
feat!: lnd on postgres

sandipndev:
This PR adds lnd on postgresql and makes it default.
Will break existing deployments without migration.

### DIFF
--- a/charts/lnd/values.yaml
+++ b/charts/lnd/values.yaml
@@ -103,3 +103,11 @@ rbac:
   create: true
 autoGenerateSeed:
   enabled: false
+
+postgresql:
+  enabled: true
+  uri: postgres://test:test@lnd-postgresql:5432/lnd
+  auth:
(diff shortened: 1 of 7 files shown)


charts/lnd/values.yaml
```diff
@@ -103,3 +103,14 @@ rbac:
   create: true
 autoGenerateSeed:
   enabled: false
+
+postgresql:
+  # Enable LND on Postgresql, postgresql.uri becomes mandatory
+  enabled: true
+  # Deploy Postgresql with this chart inside Kubernetes
+  deploy: true
+  uri: postgres://test:test@lnd-postgresql:5432/lnd
```
sandipndev, line 112: `postgresql.uri` needs to be given as a helm values input because it has to be written in configmap, is this a good practice?

bodymindarts (approved):

PR #1060, opened 2022-06-07
refactor: make testnet default for bitcoin charts

sandipndev:
Also refactors out the ports.


openoms:
Eapecially after to today's block storm on testnet I am inclined to use signet forbself-hosted experimentation.
Does not need to be covered in this PR, but would like to mention so this can be extended.

Issue #1078, opened 2022-06-09
Running the dev environment


openoms:
Running `make deploy-services` these errors thew me off:
```
╷
│ Error: Post "http://localhost/api/v1/namespaces": dial tcp [::1]:80: connect: connection refused
│ 
│   with module.infra_services.kubernetes_namespace.ingress,
│   on .terraform/modules/infra_services/modules/services/ingress.tf line 1, in resource "kubernetes_namespace" "ingress":
│    1: resource "kubernetes_namespace" "ingress" {
│ 
╵
╷
│ Error: Post "http://localhost/api/v1/namespaces": dial tcp [::1]:80: connect: connection refused
│ 
│   with module.infra_services.kubernetes_namespace.otel,
│   on .terraform/modules/infra_services/modules/services/opentelemetry.tf line 1, in resource "kubernetes_namespace" "otel":
│    1: resource "kubernetes_namespace" "otel" {
│ 
╵
```

It was simply solved by using:
```
export KUBE_CONFIG_PATH=~/.kube/config
```

Should include in the readme to be the part of the dev environment.
Working on a little script to set up the dev environment on Debian / Ubuntu:
https://github.com/openoms/bitcoin-tutorials/blob/master/k8s/devenv.k3d.sh

Should I include it here or in some dedicated documentation repo?

openoms:
Now debugging this issue close to the completion of k3d deployment when running
```
$ make deploy
```
(`terraform apply`)

```
module.galoy.kubernetes_secret.bitcoinrpc_password: Creation complete after 0s [id=galoy-dev-galoy/bitcoind-rpcpassword]                                                                                                               
module.galoy.kubernetes_secret.price_history_postgres_creds: Creation complete after 0s [id=galoy-dev-galoy/galoy-price-history-postgres-creds]                                                                                        
module.galoy.kubernetes_secret.mongodb_creds: Creation complete after 0s [id=galoy-dev-galoy/galoy-mongodb]                                                                                                                            
module.galoy.kubernetes_secret.lnd1_pubkey: Creation complete after 0s [id=galoy-dev-galoy/lnd1-pubkey]            
module.galoy.kubernetes_secret.lnd2_credentials: Creation complete after 0s [id=galoy-dev-galoy/lnd2-credentials]  
module.galoy.kubernetes_secret.lnd2_pubkey: Creation complete after 0s [id=galoy-dev-galoy/lnd2-pubkey]            
[...]

PR #1101, opened 2022-06-13
feat!: ingress config for kubernetes >=v1.22

openoms:
as according to: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

The Ingress API needs to change when running kubernetes v1.22.0 and above causing problem in the dev environment as discussed in: #1078

Staging is running kubernetes `1.21.12`
Inc is running `v1.21.11`
so the changes are compatible with them (networking.k8s.io/v1 API version, available since v1.19.)

### DIFF
--- a/dev/addons/web-wallet.tf
+++ b/dev/addons/web-wallet.tf
@@ -13,12 +13,31 @@ resource "kubernetes_secret" "web_wallet_secret" {
   }
 }
 
+resource "kubernetes_secret" "web_wallet_mobile_secret" {
+  metadata {
+    name      = "web-wallet-mobile"
+    namespace = kubernetes_namespace.addons.metadata[0].name
+  }
+
+  data = {
+    "session-keys" : local.session_keys
+  }
+}
+
 resource "helm_release" "web_wallet" {
   name      = "web-wallet"
   chart     = "${path.module}/../../charts/web-wallet"
   namespace = kubernetes_namespace.addons.metadata[0].name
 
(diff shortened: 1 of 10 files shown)


dev/addons/web-wallet.tf
```diff
@@ -19,14 +19,6 @@ resource "helm_release" "web_wallet" {
   namespace = kubernetes_namespace.addons.metadata[0].name
 
   depends_on = [kubernetes_secret.web_wallet_secret]
-}
```
krtk6160, line 22: I think we want to keep both the helm releases. One is for the web wallet desktop layout, other one for the web wallet mobile layout. I think if we rename the web wallet mobile layout helm release, it should prevent the "already exists" error that you referenced [here](https://github.com/GaloyMoney/charts/issues/1078#issuecomment-1155065631).

openoms: I have started to experiment with renaming, but then needed a new, matching named secret as well. Not sure if all be working well with 2 web-wallet servers connecetd to the same backend. Testing it is out of the scope of this PR I think. 

The `mobileLayout` is disabled in the values as well:
https://github.com/GaloyMoney/charts/blob/3f09cc0911720a0af01d48d28c95b534e1687222/dev/addons/web-wallet-mobile-values.yml#L2

i think this can be changed and applied quickly for testing.

openoms: Running another build just in case it works with a simple change.

krtk6160 (approved):

PR #1124, opened 2022-06-20
chore: set lnd log level to info

krtk6160:


nicolasburtey (approved):

charts/lnd/values.yaml
```diff
@@ -86,7 +86,7 @@ lndGeneralConfig:
   - maxchansize=500000000
   - bitcoin.timelockdelta=60
   - default-remote-max-htlcs=50
-  - debuglevel=warn
+  - debuglevel=info
```
dolcalmi, line 89: if we really want enable info for db compact then we should try something like this:
```suggestion
  - debuglevel=warn,CHDB=info
```
you can get more subsystems with `lncli debuglevel --show`

dolcalmi (changes requested):

dolcalmi (approved):

PR #1135, opened 2022-06-21
fix: add required missing properties to test accounts yaml

vindard:
## Description

After adding https://github.com/GaloyMoney/galoy/pull/1389 this broke because the original code would merge in existing array elements from `default.yaml` while the new code replaces the array entirely.

These changes are to populate the new array that would otherwise be generated from original merge code and to include that entire array here.

### DIFF
--- a/ci/testflight/galoy/testflight-values.yml
+++ b/ci/testflight/galoy/testflight-values.yml
@@ -7,22 +7,43 @@ needFirebaseServiceAccount: false
 
 galoy:
   test_accounts:
-  - phone: "+50365055543"
-    code: "182731"
-    currency: "BTC"
-  - phone: "+50365055544"
-    code: "840172"
-    currency: "BTC"
-  - phone: "+50365055545"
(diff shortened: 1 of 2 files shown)


krtk6160 (approved):

ci/testflight/galoy/testflight-values.yml
```diff
@@ -7,22 +7,43 @@ needFirebaseServiceAccount: false
...
+      phone: "+59981730222" # bankowner
+      code: "824197"
+      needUsdWallet: true
+      role: bankowner
+      username: bankowner
+    - ref: E
+      phone: "+59981730223" # dealer
+      code: "824199"
+      role: dealer
+    - ref: F
```
vindard, line 36: Original object only had 5 elements configured here, but additional 3 come from `default.yaml`. Tbd if these additional 3 are necessary for testflight.

krtk6160 (approved):

PR #1193, opened 2022-07-05
refactor: galoy chart

sandipndev:
This PR refactors the templates and brings out a good structure of default `values.yml`.

Inspired by [Concourse Helm Charts](https://github.com/concourse/concourse-chart) pattern.


sandipndev:
In order to preserve the spaces and inline comments of `values.yaml` we need to find a solution where `yq` doesn't remove them: https://github.com/GaloyMoney/galoy/blob/main/ci/tasks/bump-image-digest.sh#L12-L15

sandipndev:
For both the image refs to be updated, https://github.com/GaloyMoney/galoy/pull/1485 should go through

bodymindarts (approved):

bodymindarts (approved):

PR #1197, opened 2022-07-05
feat: use lightninglabs/loopserver for loop

openoms:
Using the new Loop image supplied by Ligthning Labs as in https://github.com/GaloyMoney/galoy/pull/1396/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3=


sandipndev (approved):

openoms:
thank you for the details @ntheile! realised that it is the wrong image  and fixed with #1199

PR #1307, opened 2022-07-29
fix: disable pg persistence on dealer testflight

krtk6160:
The postgresql chart was refactored and thus the key for persistence was moved under a new `primary` key. This PR should bring our testflight-values.yml in line with what the latest chart expects.


bodymindarts (approved):

sandipndev (approved):
Great catch!

PR #1389, opened 2022-08-12
refactor!: galoy chart with better structure

sandipndev:
Brings in a much better structure for the galoy chart.
Continuing the work in https://github.com/GaloyMoney/charts/pull/1193


sandipndev:
This PR also removes `jwt-secret` from being managed and auto-deployed via Helm.

Because of the existing annotation:
```
helm.sh/resource-policy: keep
```
It won't be deleted in existing environments but on new environments, it must be created explicitly.

charts/lnd/Chart.lock
```diff
@@ -11,5 +11,5 @@ dependencies:
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
   version: 11.1.19
-digest: sha256:6a4a8473f85ff149591396c0e2ebe47339855f1846f5fb7f6dc196e660aba1e6
-generated: "2022-08-15T21:08:35.497533254Z"
+digest: sha256:2d67304a83842f5625fd0e012d6a3e9379d1f1741723c6b288cfd72ac41d7fea
```
openoms, line 14: this is downgrading the lnd chart. Should this change be removed?

sandipndev: Yup you're right, this is happening when terraform init runs via dev imo, I'll remove it.

charts/galoy/templates/_helpers.tpl
```diff
@@ -1,16 +1,274 @@
...
+Balance Notification CronJob name
+*/}}
+{{- define "galoy.balance_notification_cron.jobname" -}}
+{{- printf "%s-balance-notif" .Release.Name -}}
+{{- end -}}
+
+{{/*
+Mongo Backup CronJob name
+*/}}
+{{- define "galoy.mongo_backup_cron.jobname" -}}
```
krtk6160, line 58: Should probably stick to a case-convention for these keys. It's snake case here and kebab case on line 72 - `pre-migration`.
I've only seen camelcase being the convention in most helm charts, so we should think about this.

charts/galoy/templates/galoy-serviceaccount-rbac.yaml
```diff
@@ -0,0 +1,41 @@
...
+  resources: ["cronjobs"]
+  verbs: ["get"]
+- apiGroups: ["batch"]
+  resources: ["jobs"]
+  verbs: ["create", "delete", "get", "list"]
+
+---
+
+apiVersion: {{ include "common.capabilities.rbac.apiVersion" . }}
+kind: RoleBinding
```
krtk6160, line 31: Any reason for putting 3 different resource types put into the same file?

charts/galoy/values.yaml
```diff
@@ -1,106 +1,576 @@
...
+      ## Backup data to Google Cloud Storage Bucket
+      ## To backup to GCS bucket, you would need a GCP Service Account
+      ## that lets you upload to a specific bucket
+      ##
+      gcs:
+        # Enable/disable uploading backup to GCS bucket
+        enabled: false
+
+        # Configuration for GCS bucket
+        config:
```
krtk6160, line 238: I don't think we need a `config` key for the gcs or dropbox backups. It's adding an additional level of nesting, which imo isn't needed

krtk6160 (commented):
Have made a first round of comments

charts/galoy/values.yaml
sandipndev: We need this to specify the GCS bucket name

charts/galoy/templates/galoy-serviceaccount-rbac.yaml
sandipndev: No reason, will split them out

charts/galoy/values.yaml
krtk6160: why can't we specify the gcs bucket name like:
```
gcs:
  enabled: true
  bucketName: xyz
```

sandipndev: Ah, okay I will incorporate this.

sandipndev: Done

bodymindarts (approved):

PR #1404, opened 2022-08-12
ci: run ga dev smoketest sequencial jobs

sandipndev:


sandipndev:
No wait this doesn't work

PR #1468, opened 2022-08-19
fix: helm context for bitcoind prometheus port

zoop-btc:
Just a small error I've noticed. The context is wrong thus helm would fail when supplying this with a values file.


bodymindarts (approved):

krtk6160 (approved):
LGTM

PR #1521, opened 2022-08-26
chore: remove fake galoyapi from dealer chart

sandipndev:
Removes fake galoyapi now that dealer talks with real galoy graphql api


openoms:
Ran without issues with the local dev test and there was no fake-galoyapi coming up.

openoms (approved):

Issue #1240, opened 2022-07-15
implement remote signing


nicolasburtey:
It's possible to "segregate" lnd in 2, where one lnd have all the states, and another lnd will just be signer. more info [here](https://github.com/lightningnetwork/lnd/blob/master/docs/remote-signing.md). 

This would allow to have the signer in a different cluster/environement, with more stringent access. 

Also, maybe this provide the ability for a "kill switch", just by removing network to the `signer` lnd.

Note: lnd 15 seems to have a regression on this (at the time of wallet creation only), so better to test with lnd 14.3 or wait for a patch lnd 15.1

openoms:
Depends on https://github.com/lightningnetwork/lnd/pull/7056 to avoid scanning the full chain on wallet creation.

nicolasburtey:
> 

how long does it take on signet and mainnet?

openoms:
On mainnet it is multiple hours (~2days on my test RPi4) due to scanning from the genesis (not even the first SegWit block).

On signet it is minutes because the chain is much shorter and there are only a few transactions.

openoms:
Related PRs in this repo:
* remove lndmon:  
#2139
* move loop to a standalone chart:  
#2163
* duplicate the lnd chart to an `lnd-segregated` chart in which:
  * run either the lnd-signer or lnd-watchonly config
  * connect to loop
  * remove the postgres backend related code
  #1581

PR #1581, opened 2022-09-12
feat: implement lnd remote signing

openoms:
Fixing: https://github.com/GaloyMoney/charts/issues/1240

added lndsigner as a subchart under lnd
it is the copy of the lnd  charts except:
vars are named lndsigner
removed:
P2P port, REST port, Tor connection

config applied as in https://github.com/lightningnetwork/lnd/blob/master/docs/remote-signing.md#the-signer-node

Values to be aware of:
LNDSIGNER_PASS
 lndsigner-password

The watch only node will node the configured as in:
https://github.com/lightningnetwork/lnd/blob/master/docs/remote-signing.md#the-watch-only-node
and needs `remotesigner.migrate-wallet-to-watch-only=true ` if migrating an existing node (with added manual steps TBD).


openoms:
Rebased to the latest main.

Successfully created a watch-only lnd instance on regtest with the settings changed in `lnd-regtest-values.yaml`:
```
lndWatchOnly:
  enabled: true
  signer: signer1.galoy-dev-signer.svc.cluster.local:10009

autoGenerateSeed:
  enabled: false
```

For now lnd needs to be stopped and manually initialized - once only.
This can be achieved with inserting the following line:
```
        while true; do echo "$date sleeping an hour"; sleep 3600; done
```
before lnd runs: https://github.com/GaloyMoney/charts/blob/6a71f8c1ba5f34ebf144a6ef52930a8b9f9b58bf/charts/lnd/templates/statefulset.yaml#L162

Then: open two terminals with the lnd pod with:
```
$ k -n galoy-dev-bitcoin exec -it    lnd1-0 -c lnd -- sh
```
in terminal 1 run lnd:
```
$ lnd
```
in terminal 2 create the wallet with:
```
$ lncli -n regtest createwatchonly /root/.lnd/accounts-signer.json                      
Input wallet password:                                                                    
Confirm password:                                                                         
                                             
Input an optional wallet birthday unix timestamp of first block to start scanning from (default 0): 
                                             
                                             
Input an optional address look-ahead used to scan for used keys (default 2500): 
              
```
for just testing purposes can use any min 8character password locally and accept the default timestamp and address depth.

To see if successful:
```
$ lncli -n regtest state
{
    "state": "SERVER_ACTIVE"
}
```

Should be able to automate the watch-only wallet creation as well later.

Currently proceeding to signet to be ready for the manual staging deployment.

openoms:
Added the expect script for creating the watch only wallet as well so the wallet can be initialized automatically when set.
For the signer creating the seed is deafult in the dev setup:
```
autoGenerateSeed:
  enabled: true
``` 
For lnd the settings to enable watch only and create the wallet from the signer accounts are:
```
  lndWatchOnly:
  enabled: true
  signer: signer1.galoy-dev-signer.svc.cluster.local:10009

autoGenerateSeed:
  enabled: true
```

Tested on signet also which took ~90 minutes to scan the whole chain from genesis (as currently lnd does it until the fix is merged and released: https://github.com/lightningnetwork/lnd/pull/7056)

charts/lnd/templates/statefulset.yaml
```diff
@@ -153,6 +158,14 @@ spec:
             mountPath: /root/.lnd
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          {{- if .Values.lndWatchOnly.enabled }}
+          command:
```
krtk6160, line 162: We might not need a custom command to base64 decode the macaroon if we skip base64 encoding it before passing it to `kubectl create secret`.

charts/lnd/templates/statefulset.yaml
```diff
@@ -226,11 +250,31 @@ spec:
...
-        {{- if ne .Values.global.network "regtest" }}
+        {{- if or (ne .Values.global.network "regtest") .Values.lndWatchOnly.enabled }}
         - name: lnd-password
           secret:
             secretName: {{ printf "%s-pass" (include "lnd.fullname" .) }}
         {{- end }}
+        {{- if .Values.lndWatchOnly.enabled }}
+        - name: signer-tls
+          secret:
+            secretName: signer1-credentials
```
krtk6160, line 261: Why is there a "1" suffix for "signer" in the secret name?

dev/signer/signer1.tf
```diff
@@ -0,0 +1,11 @@
+resource "helm_release" "signer" {
```
krtk6160, line 1: Should probably keep all of this in `signer/main.tf`. The size of `signer/main.tf` with this content isn't so big (yet) that it would warrant a split into a new file.

krtk6160 (commented):
A first round of comments

charts/lnd/templates/statefulset.yaml
openoms: I found the same problem when exporting the `signer.custom.macaroon` to the secrets from the file. 
It did not decode to the same thing when decoded to be mounted in the pod so using the base64 encoded macaroon here.

This decode step is using the secret which was created from:
```
export MACAROON=$(base64 /root/.lnd/data/chain/bitcoin/{{.Values.global.network}}/signer.custom.macaroon | tr -d '\n\r')
 ```
I think this already solves the problem you mentioned and tested ok.

File: charts/lnd/templates/statefulset.yaml (hunk 2 above)
Comment (openoms) [reply charts/lnd/templates/statefulset.yaml:261]: Similar to the full featured LND nodes the signer nodes can be two as well paired with the watch-only lnd-s.
If we run two nodes in the cluster the same distinction is needed. Will be more apparent on staging.

File: dev/signer/signer1.tf (hunk 3 above)
Comment (openoms) [reply dev/signer/signer1.tf]: changed in [c575a02](https://github.com/GaloyMoney/charts/pull/1581/commits/c575a02f6365743ae3a32777ec19faac4e11c676)

File: charts/lnd/templates/statefulset.yaml (hunk 2 above)
Comment (krtk6160) [reply charts/lnd/templates/statefulset.yaml:261]: But this code is hard coding the lnd sts to only work with secrets whose name is `signer1-credentials`. If we wanted to use lnd2 too with its own remote signer, it wouldn't work, because lnd2's sts would reference `signer1-credentials` instead of `signer2-credentials`

Comment (openoms) [reply charts/lnd/templates/statefulset.yaml]: fixed in [111ff6a](https://github.com/GaloyMoney/charts/pull/1581/commits/111ff6a6bbbcc5ee2af7d0b162bf3e5fb6c5ba67)

**Comment by openoms** [comment]
Rebased to main to include the LND update and updated the signer image to LND v0.15.4.

**Review by krtk6160 — APPROVED** [review]

**Comment by openoms** [comment]
Ready for review again after the changes suggested @bodymindarts 
The `lnd-watchonly` configuration now is a completely separate chart from lnd with the different options of (as in the dev setup):

```
# sets up the watchonly wallet from the signer accounts and tls (provided in secrets)
autoCreateWatchonly:
  enabled: true

signerConfig:
  signerHost: lnd-signer1.galoy-dev-signer.svc.cluster.local:10009
  signerCredentialsSecretName: lnd-signer1-credentials
  signerAccountsSecretName: lnd-signer1-accounts
```

The signer chart is renamed to `lnd-signer` and has the option:
```
autoGenerateSeed:
  enabled: true
```
similar to the lnd chart.

The dev setup is currently configured to set up an `lnd-watchonly` - `lnd-signer` pair on regtest.

bodymindarts (commented):
I think lnd-watchonly has a bunch of sub charts that we can get rid of or? loop, loop server, lndmon etc?

openoms:
> I think lnd-watchonly has a bunch of sub charts that we can get rid of or? loop, loop server, lndmon etc?

lnd-watchonly would run in place of the full lnd so to have those functions working those subcharts are needed. Also it is tested to run ok on regtest.

openoms:
The current state:

The `lnd-segregated` chart is fully functional in two parts:
 * `mode: watchonly` 
   * runs lnd without the keys
   * exposes the same macaroon + tls.cert authenticated GRPC and REST interface as the standalone lnd chart
   * connects to bitcoind and watches the chain
   * can set `autoCreateWatchonly` which imports the provided accounts and connects to the signer
 * `mode: signer` 
   * runs the signer only with the keys
   * exposes only a signer-macaroon + signer-tls.cert authenticated GRPC endpoint internally
   * there is no connection to bitcoind (blind)
   * can set `autoGenerateSeed` to create the keys to test

There are no subcharts.

Tested in the dev setup and in github actions as `lnd1` (unchanged standalone lnd) and lnd2: `lnd2-watchonly` connected to the `lnd-signer2`.

openoms:
The last changes introduce  the `mode: complete` in the lnd-segregated chart values which makes it able to stand up as a single helm release.

Tested locally in Actions as lnd1 in the dev setup.

The development of this feature pauses here as currently not needed in production despite in a functional state. Can be continued / taken over as needed.

openoms:
closing now to reopen if to be implemented

PR #1596, opened 2022-09-14
chore: use bbolt by default for lnd 

openoms:
all production and staging is running with bbolt so changing the dev env to bbolt too

### DIFF
--- a/charts/lnd/values.yaml
+++ b/charts/lnd/values.yaml
@@ -101,13 +101,13 @@ rbac:
 autoGenerateSeed:
   enabled: false
 postgresql:
-  enabled: true
+  enabled: false
   auth:
     database: lnd
     existingSecret: postgres-creds
 lnd:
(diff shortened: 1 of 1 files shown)


krtk6160:
@openoms I'm thinking maybe we should make bbolt default in the chart itself, given the flaky nature of lnd on postgres.

charts/lnd/values.yaml
```diff
@@ -101,13 +101,13 @@ rbac:
...
   auth:
     database: lnd
     existingSecret: postgres-creds
 lnd:
   db:
-    backend: postgres
+    backend: bbolt
     config:
       # This secret should contain `uri` as the key
       secret: postgres-creds
```
openoms, line 113: @sandipndev I wonder if it is useful to keep these postgres-creds entries in case we would want to use postgres again

PR #1645, opened 2022-09-20
chore: use tracing prefix for TRACING_SERVICE_NAME

sandipndev:
Brings back tracing name prefix

### DIFF
--- a/charts/galoy/values.yaml
+++ b/charts/galoy/values.yaml
@@ -560,7 +560,7 @@ tracing:
   jaegerHost: localhost
   ## Prefix for tracing
   ##
-  prefix: galoy-staging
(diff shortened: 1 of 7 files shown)


charts/galoy/values.yaml
```diff
@@ -558,6 +558,9 @@ tracing:
   ## Jaeger Tracing Host
   ##
   jaegerHost: localhost
+  ## Prefix for tracing
+  ##
+  prefix: galoy-staging
```
bodymindarts, line 563: I think the default `name-prefix` should be `galoy-dev`
```suggestion
  prefix: galoy-dev
```

bodymindarts (approved):

PR #1657, opened 2022-09-21
fix: fix web wallet smoketest

Extheoisah:


ci/tasks/web-wallet-smoketest.sh
```diff
@@ -13,7 +13,8 @@ for i in {1..15}; do
   echo "Attempt ${i} to curl web wallet"
   curl --location -f ${host}:${port}
   status=$?
-  curl --location -f ${web_wallet_mobile_host}:${port}
+  if [[ web_wallet_mobile_host == "" ]]; then echo "No web wallet secondary host available";
```
bodymindarts, line 16: This reads like its a negative even though it is expected.
```suggestion
  if [[ web_wallet_mobile_host == "" ]]; then echo "Skipping secondary host - not configured."; 
```

bodymindarts: I think you've got the syntax wrong:
```
if [[ .. ]]; then
  true thing
 else
   false thing
 fi
```

Extheoisah: @bodymindarts Like this right
```
if [[ web_wallet_mobile_host != "" ]]; then
    curl --location -f ${web_wallet_mobile_host}:${port};
  else
    echo "Skipping secondary host - not configured.";
  fi
  ```

bodymindarts: You shouldn't need the `;` at the end of the lines. And formatting (indentation) is off.

PR #1670, opened 2022-09-23
ci: github action to determine chart.lock out of sync issues

sandipndev:


krtk6160:
@sandipndev is this still relevant?

sandipndev:
Yup, it’ll let you know if Chart.lock is outdated with Chart.yaml

We can see failures for it when we deploy the testflight environment. This will help us understand it in advance.

Just need to check the code though if it still does what I think it does. Will get back.

krtk6160:
Closing this as I'm not sure if it's still relevant. Feel free to reopen.

PR #1791, opened 2022-10-10
feat: add RTL to the dev setup

openoms:
Adding RTL to the local dev deployment.

To access RTL on the local regtest:
```
env=galoy-dev

# get the password
k -n ${env}-bitcoin get secrets rtl-pass -o jsonpath='{.data.password}' | base64 -d

# forward the port from the pod
k -n ${env}-bitcoin port-forward  $(kubectl get pods -A | grep rtl | awk '{print $2}')  3000:3000

# access RTL on your http://localhost:3000
```

Next step is to update update to the [latest version](https://hub.docker.com/layers/shahanafarooqui/rtl/0.13.1/images/sha256-1d2892e29b5ccc4e9768ce0e7764bc9a1fd46c82d02131204f22a6aa239e106b?context=explore),but it doesn't work yet out of the box due to configuration changes.

### DIFF
--- a/dev/bitcoin/rtl.tf
+++ b/dev/bitcoin/rtl.tf
@@ -1,12 +1,9 @@
 resource "helm_release" "rtl" {
-  name       = "rtl"
-  chart      = "${path.module}/../../charts/rtl"
-  namespace  = kubernetes_namespace.bitcoin.metadata[0].name
+  name      = "rtl"
+  chart     = "${path.module}/../../charts/rtl"
+  namespace = kubernetes_namespace.bitcoin.metadata[0].name
 
   dependency_update = true
-  values = [
-    file("${path.module}/rtl-values.yml")
-  ]
 
   depends_on = [
     helm_release.lnd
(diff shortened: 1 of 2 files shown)


dev/bitcoin/rtl.tf
```diff
@@ -0,0 +1,14 @@
+resource "helm_release" "rtl" {
+  name       = "rtl"
+  chart      = "${path.module}/../../charts/rtl"
+  namespace  = kubernetes_namespace.bitcoin.metadata[0].name
+
+  dependency_update = true
+  values = [
+    file("${path.module}/rtl-values.yml")
```
krtk6160, line 8: Can probably get rid of this since it's empty.

openoms: cleaned out in [8e54598](https://github.com/GaloyMoney/charts/pull/1791/commits/8e5459805ad4e9fbdfadc00a6c9a5bda26e4b662)

krtk6160 (approved):

PR #1849, opened 2022-10-18
chore: remove galoy-auth chart and introduce kratos to galoy chart

sandipndev:
This PR deprecates `galoy-auth` chart and makes kratos dependency of galoy chart.


bodymindarts:
This can not be 1 PR as in inc and staging we are already using the `galoy-auth` kratos. We must first deploy the `galoy` lifecycle kratos, then migrate, then delete `galoy-auth`

PR #1859, opened 2022-10-19
chore: Increased proxy timeouts for websockets

daviroo:
I found some docs on websockets config for kubernetes nginx - it says we should have a higher value for some configs for proxy for websockets.

https://kubernetes.github.io/ingress-nginx/user-guide/miscellaneous/#websockets

### DIFF
--- a/charts/galoy/templates/api-ingress.yaml
+++ b/charts/galoy/templates/api-ingress.yaml
@@ -13,6 +13,9 @@ metadata:
   annotations:
     cert-manager.io/cluster-issuer: {{ .Values.galoy.api.ingress.clusterIssuer }}
 
+    nginx.ingress.kubernetes.io/proxy-read-timeout: "21600" # 6 hours
+    nginx.ingress.kubernetes.io/proxy-send-timeout: "21600" # 6 hours
+
     nginx.ingress.kubernetes.io/limit-rpm: "60"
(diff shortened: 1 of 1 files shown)


charts/galoy/templates/api-ingress.yaml
```diff
@@ -13,6 +13,9 @@ metadata:
   annotations:
     cert-manager.io/cluster-issuer: {{ .Values.galoy.api.ingress.clusterIssuer }}
 
+    nginx.ingress.kubernetes.io/proxy-read-timeout: "21600" # 6 hours
```
nicolasburtey, line 16: is this not missing the unit?

https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/advanced-configuration-with-annotations/

if default is 60s, not sure we really that big of a value? maybe we could try 10 min and see how that works. I think eventually those value should only apply to the websocket, not the public api for normal http request

nicolasburtey: let's try 1h then? 

it should also be seen in regards to our usage.  I'm doubting much sessions last 1h+ in our case. 

longer session make it prone to some attack (like [slowloris](https://www.cloudflare.com/learning/ddos/ddos-attack-tools/slowloris/)) so better to not be overly large here

nicolasburtey (approved):

PR #1866, opened 2022-10-20
feat: add kratos master password

sandipndev:


nicolasburtey (approved):

nicolasburtey:
not generated via terraform?

sandipndev:
> not generated via terraform?

It will be at deployments. This is the injection of the environment variable via helm.

PR #1873, opened 2022-10-21
chore: Increased connections limit for api ingress

daviroo:
I can see in the logs that we are actually rate limiting based on number of connections coming from an IP.  I've doubled it for now to see what effect it will have (previously 40 now 80).  This doesn't effect the other rate limits we have in effect.


nicolasburtey:
`40` seems pretty high already?

it would basically mean 40 websockets session before it hits the limits? assuming query and mutations takes ms to return

nicolasburtey:
approved so we can see the effect but we should revert if we don't see any change

nicolasburtey (approved):

PR #1912, opened 2022-10-26
fix: fill loop creds with lnd data

openoms:
When using sigent Loop needs to be disabled, but the galoy API needs it's credentials for now.

This is a temporary solution as did not get to avoid  the errors in the backend yet: https://github.com/GaloyMoney/galoy/pull/1733 only discussed steps needed there with @nicktee.


openoms:
Looking into fix this in the galoy repo to avoid needing to pass fake values.

PR #2077, opened 2022-11-28
feat: swap fee accounting config

ntheile:


bodymindarts:
Why the different values in signet / and regtest? Which one is the default?

ntheile:
The default in the source code is `true`: https://github.com/GaloyMoney/galoy/blob/main/galoy.yaml#L182 

But in the deployment it is set to `false`: https://github.com/GaloyMoney/galoy-deployments/pull/2644/files#diff-f18d77850a07073c181f62369beae2503b63697286ce8caa4cac7c7f67446a72R77 

I am unsure in the local dev charts the order things get applied, so to be sure I just explicitly set it.

bodymindarts:
If the default is `true` then I'd say you should only override it in signet. Only one of the files gets applied - not both.

nicolasburtey (approved):

PR #2124, opened 2022-12-07
chore: move ingress-nginx to galoy-deps

openoms:
Moved over the ingress charts and config from galoy-infra, but in the current state getting hit by the error:
```
╷                             
│ Error: Failed to determine GroupVersionResource for manifest
│                                       
│   with module.galoy_deps.kubernetes_manifest.issuer,
│   on galoy-deps/ingress.tf line 47, in resource "kubernetes_manifest" "issuer":
│   47: resource "kubernetes_manifest" "issuer" {    
│                                                                                         
│ no matches for kind "ClusterIssuer" in group "cert-manager.io"
╵                                                                                         
make: *** [Makefile:18: deploy-services] Error 1   
```


krtk6160:
> Moved over the ingress charts and config from galoy-infra, but in the current state getting hit by the error:
> 
> ```
> ╷                             
> │ Error: Failed to determine GroupVersionResource for manifest
> │                                       
> │   with module.galoy_deps.kubernetes_manifest.issuer,
> │   on galoy-deps/ingress.tf line 47, in resource "kubernetes_manifest" "issuer":
> │   47: resource "kubernetes_manifest" "issuer" {    
> │                                                                                         
> │ no matches for kind "ClusterIssuer" in group "cert-manager.io"
> ╵                                                                                         
> make: *** [Makefile:18: deploy-services] Error 1   
> ```

Right that's because the ingress chart uses a CRD. Because of the way the provider works, we cannot do the installation in one go, we need to first install the chart (which will install the CRD) and then the CR.

openoms:
The problem still persists, but there are two kind of hacky solutions suggested here: https://stackoverflow.com/questions/69765121/how-to-avoid-clusterissuer-dependency-on-helm-cert-manager-crds-in-terraform-pla

dev/main.tf
```diff
@@ -2,21 +2,9 @@ variable "bitcoin_network" {}
...
 
-module "infra_services" {
-  source = "git::https://github.com/GaloyMoney/galoy-infra.git//modules/services?ref=86b0906"
-
-  name_prefix                 = local.name_prefix
-  letsencrypt_issuer_email    = local.letsencrypt_issuer_email
-  local_deploy                = true
-  cluster_endpoint            = "dummy"
-  cluster_ca_cert             = "dummy"
-  honeycomb_api_key           = "dummy"
```
ntheile, line 18: @openoms i am trying to configure my local dev charts environment to use my honeycomb api keys. Is there a new way to do this now?

openoms: The setting is here now:
https://github.com/GaloyMoney/charts/blob/d526361517dd559c7669a0b210b5842666e88b6c/charts/galoy-deps/values.yaml#L23
can you try filling that?

PR #2139, opened 2022-12-12
chore: remove lndmon, test lnd_api_endpoint and lnd_p2p_endpoint

openoms:


ci/tasks/lnd-smoketest.sh
```diff
@@ -4,20 +4,17 @@ set -eu
...
 lnd_host=`setting "lnd_p2p_endpoint"`
 
-if [[ "${lndmon_host}" != "" ]]; then
+if [[ "${lnd_host}" != "" ]]; then
   set +e
   for i in {1..60}; do
-    echo "Attempt ${i} to curl lndmon"
-    curl -f ${lndmon_host}:9092/metrics
+    echo "Attempt ${i} to connect the lnd_p2p_endpoint"
+    nc -zv ${lnd_host} 9735
```
krtk6160, line 13: Is checking the p2p port connectivity equivalent to the previous smoketest?

Which port does lndmon use to connect to lnd to gather metrics? iirc it was the rpc port. If so, then we should also test connectivity to the rpc port as our other services directly rely on that.

krtk6160 (commented):
I have a comment regarding the smoketest, otherwise LGTM

ci/tasks/lnd-smoketest.sh
openoms: In the lndmon chart the GRPC port was hardcoded as: `--lnd.host={{.Release.Name}}:10009` so changed to test that first and proceed to the p2p endpoint.

krtk6160 (approved):

anthr76:
Hello

Is it possible to have some context on this change? It looks like to me lndmon would be removing thus loosing metrics to lnd. Is it planned to break out lndmon to another helm chart?

openoms:
Hi @anthr76. lndmon was removed for multiple reasons:
* it does not support signet (https://github.com/lightninglabs/lndmon/issues/80) where we moved our staging deployment
* the function we used can be replaced with the galoy backend: https://github.com/GaloyMoney/galoy/pull/2057
* there is less infrastructure to manage and LND now has no subcharts

Currently it is not planned to reintroduce lndmon as a standalone chart, but could be an option as it was done with loop also.

PR #2163, opened 2022-12-13
chore: move loop to a standalone chart

openoms:
The aim of this PR to move the Loop chart out from underneath LND and be a standalone chart. Loopserver remains a subchart of Loop for regtest only.
Loop does not run on signet and tested to not come up.

This is the next step towards the duplication of the LND chart to have a second segregated LND chart available with signer and watchonly instances.
If Loop would be a subchart it would need to be also duplicated and maintained in two places with the same function.

 Renaming the secrets in line with the Loop release name change.
 Not changing the galoy environment variables:
`LND1_LOOP_MACAROON`
`LND1_LOOP_TLS`
`LND2_LOOP_MACAROON`
`LND2_LOOP_TLS`


krtk6160 (approved):

openoms:
recovered some changes after the rebase. The local test is successful. Waiting for the green here.

krtk6160 (approved):

ntheile:
@openoms just saw this, will test and review tomorrow

openoms:
> @openoms just saw this, will test and review tomorrow

Nick, just saw your comment as I pressed on merge. Hope it will not disturb your work much.

PR #2186, opened 2022-12-17
feat: export lnd's TLS secrets as a k8s tls secret type

anthr76:
The idea behind this is to have the ability to plug in the certificates to something like your ingress-controller which typically requires a standard `kubernetes.io/tls` secret type.

I do hope to hopefully manage certificates outside of LND once https://github.com/lightningnetwork/lnd/pull/6479 is implemented.

Signed-off-by: Anthony Rabbito <hello@anthonyrabbito.com>

### DIFF
--- a/charts/lnd/Chart.lock
+++ b/charts/lnd/Chart.lock
@@ -11,5 +11,5 @@ dependencies:
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
   version: 11.1.19
-digest: sha256:b4ff598f1a247a7d42db9e87c24ff35328829544e794294c12bb147d5d1efc49
-generated: "2022-11-15T17:06:55.266810345Z"
+digest: sha256:7f4c944794bcc39964a49eef6a4ecff4ceeb456957f4b0f47d2519935e7f6b78
+generated: "2022-12-17T01:59:28.469117336-05:00"
--- a/charts/lnd/Chart.yaml
+++ b/charts/lnd/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: lnd
-version: 0.3.10-dev
+version: 0.3.11-dev
 appVersion: 0.15.4
 description: LND helm chart
 keywords:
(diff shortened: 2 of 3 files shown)


bodymindarts:
Hi @anthr76,

first of all thanks for your contribution. I think it could certainly make sense to persist these certs as a `tls` secret as that is what they are - I'm even thinking we should perhaps _only_ persist them there.

I do wonder however why you would want to connect them to your ingress. We currently only have internal services talking to lnd via the private network. I would consider it a security concern to expose the api publicly.

2 questions:
- what is the specific use-case you are trying to achieve with this change?
- are you blocked currently and this PR is the only way you see achieving it?

bodymindarts:
Hi @anthr76,

first of all thanks for your contribution. I think it could certainly make sense to persist these certs as a `tls` secret as that is what they are - I'm even thinking we should perhaps _only_ persist them there.

I do wonder however why you would want to connect them to your ingress. We currently only have internal services talking to lnd via the private network. I would consider it a security concern to expose the api publicly.

2 questions:
- what is the specific use-case you are trying to achieve with this change?
- are you blocked currently and this PR is the only way you see achieving it?

anthr76:
> first of all thanks for your contribution. I think it could certainly make sense to persist these certs as a tls secret as that is what they are - I'm even thinking we should perhaps only persist them there.

Agreed. I'm happy to look where else they are use in the lnd-credentials secret and make use of the new secret.

> I do wonder however why you would want to connect them to your ingress. We currently only have internal services talking to lnd via the private network. I would consider it a security concern to expose the api publicly.

Sure! So to preface I'm quite new to lightning so I'm happy to stand corrected anywhere. In fact I tried to join the telegram group to ask a few preliminary questions but I found out I needed to send a tiny amount of sats. Which was hard to do for me at the time until I stood up my infra :(

Currently I'm not (or at least yet) making use of GaloyMoney, as I'm looking to use lnd and bitcoind to use things like Podcast2.0 and Alby. Since I have a quite fault tolerant on prem Kubernetes clusters these charts seemed to implement that in the best way. Sadly though, a lot of things (like alby) expect the API server to be exposed to the internet.

If you do find this is a high risk I will happily roll with your guidance and take the implications of keeping the service internal only and using VPNs. I figured considering the macaroons worked like a JWT key as long as that is not leaked out I'm safer... or at least safer. In fact I'd like to purely use tor if that's possible and avoid clearnet.  

> 2 questions:
> 
> what is the specific use-case you are trying to achieve with this change?

I think this was quite answered up above.

> are you blocked currently and this PR is the only way you see achieving it?

Not now. My work around was just to expose Rest API with a Let'sEncrypt cert fronting the lnd's certificates. This allows me to use Alby and Zues.

anthr76:
Closing this. Feel free to poke if we want to merge this

PR #2200, opened 2022-12-19
chore!: remove pg from lnd

krtk6160:
LND on postgres is far from production ready. We want to keep the chart clean and only allow configurations that run well in prod, so this PR removes postgres support from the chart.


openoms:
Local test was running ok.

openoms (approved):

PR #2218, opened 2022-12-21
fix: disable loop for testflight on signet

openoms:
~~Fix the loop secret names after changed in https://github.com/GaloyMoney/charts/pull/2163~~

### DIFF
--- a/ci/testflight/galoy/testflight-values.yml
+++ b/ci/testflight/galoy/testflight-values.yml
@@ -75,3 +75,6 @@ oathkeeper:
 
 secrets:
   create: false
+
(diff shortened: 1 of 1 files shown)


krtk6160 (approved):

openoms:
Since the testflight is running on signet the loop credentials won't be present. Disabled loop for galoy in the testflight with [4f92a23](https://github.com/GaloyMoney/charts/pull/2218/commits/4f92a23fd0b2e73712013324aa9ba66975db4628)

krtk6160 (approved):

PR #2279, opened 2022-12-29
chore: update rtl to v0.13.3

zoop-btc:
This updates the rtl chart to the latest version. It needed some changes since newer rtl versions need write-access to the config file.
Which is why I added an init-container which copies over the configuration and sets write permissions for the node-user.
With this the folder /RTL/data will hold the config, lnd-macaroons, lnd-channelbackups & lnd logs.


openoms:
Works ok on regtest, approving.

openoms (approved):

bodymindarts (approved):

PR #2409, opened 2023-01-19
chore: order rtl after loop in dev

openoms:


dev/Makefile
```diff
@@ -32,12 +32,12 @@ all: create-cluster init deploy-services deploy
 
 
 redeploy-bitcoin:
-	kubectl delete namespace galoy-dev-bitcoin --force
+	-kubectl delete namespace galoy-dev-bitcoin --force
```
krtk6160, line 35: What is this syntax? Never seen it before

dev/bitcoin/rtl.tf
```diff
@@ -6,6 +6,7 @@ resource "helm_release" "rtl" {
   dependency_update = true
 
   depends_on = [
-    helm_release.lnd
+    helm_release.lnd,
+    helm_release.loop
```
krtk6160, line 10: Do we need this to solve a race condition? Are we seeing the same problem here as in deployments?

krtk6160 (commented):
Have a couple of questions

dev/Makefile
openoms: It makes that command allowed to fail and the Makefile to continue. 
In case the redeploy was run and failed before the namespace is already deleted so the command is unusable and need to run the `tf apply ...` manually
a `-` prefix or `|| true` on the end allows it to continue.

dev/bitcoin/rtl.tf
openoms: Didn't have a race condition before in dev, but would be good to follow the better practice we learnt from production. Also good for testing as github actions are more similar this way to production.

krtk6160 (approved):

PR #2426, opened 2023-01-24
chore: add the kratos and oathkeeper config to the default values

openoms:
The Oathkeeper changes are based on:
https://github.com/GaloyMoney/galoy/pull/2103/files#diff-f2d3bed1938639bd048c1629be6fe575621bc4eb0cef967ab52d10f6ca96369e

Kratos:
https://github.com/GaloyMoney/galoy/blob/6f3bcdbb4bc24beb3546a5ade90a5384cc707072/dev/ory/kratos.yml

except:
`schemas`
`dsn`

A new postgresql instance is added to stand up before kratos and the `dsn` is configured accordingly.

### DIFF
--- a/charts/galoy/values.yaml
+++ b/charts/galoy/values.yaml
@@ -721,7 +721,7 @@ kratos:
 
       version: v0.11.0
 
-      dsn: memory
+      dsn: postgresql://kratos-pg:kratos-pg@postgresql/kratos-pg
 
       serve:
         public:
--- a/charts/galoy/values.yaml
+++ b/charts/galoy/values.yaml
@@ -715,23 +715,170 @@ kratos:
     automigration:
       enabled: true
     config:
-      session:
-        lifespan: "720h" # 1 month
-        earliest_possible_extend: "96h" # 4 days
-      identity:
-        schemas:
-          - id: email
-            url: file:///etc/config/identity.schema.json
-        default_schema_id: email
-      selfservice:
-        default_browser_return_url: http://127.0.0.1:4002/
-      courier:
-        smtp:
-          connection_uri: smtps://test:test@mailslurper:1025/?skip_ssl_verify=true
-      dsn: memory
       secrets:
         default:
           - PLEASE-CHANGE-ME-I-AM-VERY-INSECURE
+
+      version: v0.11.0
+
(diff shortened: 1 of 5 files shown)


krtk6160 (approved):

charts/galoy/Chart.yaml
```diff
@@ -31,6 +31,11 @@ dependencies:
     repository: https://charts.bitnami.com/bitnami
     version: 11.9.13
     condition: postgresql.enabled
+  - alias: kratospg
```
krtk6160, line 34: should be kratos-pg imo

openoms: yes, it was a problem in the templates/galoy-secrets.yaml

anyway the the kratos chart has an automigration init container which runs before the postgresql deployment in the same chatr so need to move the postgres instance out

charts/galoy/values.yaml
```diff
@@ -708,23 +715,170 @@ kratos:
...
+                    config:
+                      # url: http://e2e-tests:4002/auth/after_settings_hooks
+                      url: http://invalid-because-we-dont-want-profile-to-be-updated
+                      method: POST
+                      body: file:///home/ory/body.jsonnet
+                      auth:
+                        type: api_key
+                        config:
+                          name: Authorization
+                          value: The-Value-of-My-Key
```
ntheile, line 801: hey @openoms , i have been troubleshooting in my local dev charts environment and I think we need to read this value from the `callback_api_key` from here https://github.com/GaloyMoney/charts/blob/main/dev/galoy/main.tf#L314. I am not quite sure how to configure it here in the yaml file.

charts/galoy/values.yaml
```diff
@@ -708,23 +715,170 @@ kratos:
...
+                    config:
+                      url: http://e2e-tests:4012/kratos/registration
+                      method: POST
+                      can_interrupt: true
+                      body: file:///home/ory/body.jsonnet # TODO: use a base64 encoding instead
+                      auth:
+                        type: api_key
+                        config:
+                          name: Authorization
+                          value: The-Value-of-My-Key
```
ntheile, line 842: @openoms same here

PR #2437, opened 2023-01-25
chore: use pod ip instead of 0.0.0.0 in otel

krtk6160:
The upgrade to 0.47 uses pod ip instead of 0.0.0.0 due to security reasons and thus we should also make the switch to follow best practices.


krtk6160:
#2436 should be merged after this

openoms (approved):

PR #2571, opened 2023-02-15
chore: add defaults for galoy kratos values

openoms:
related to: https://github.com/GaloyMoney/galoy/pull/2254


openoms:
rebased on latest main to incorporate the updated oathkeeper config

PR #2603, opened 2023-02-21
chore: add admin_accounts to the galoy-smoketest

openoms:


ntheile (approved):

ntheile (approved):

ci/tasks/galoy-smoketest.sh
```diff
@@ -63,10 +67,10 @@ break_and_display_on_error_response
 set +e
 for i in {1..15}; do
   echo "Attempt ${i} to curl the admin-backend route"
-  curl -LksSf  "${host}:${port}/admin/graphql" \
+  curl -LksSf  "${admin_host}:${port}/admin/graphql" \
```
ntheile: i think we need to change this for the staging env to point to `${admin_host}:${port}/admin/graphql` maybe have an if statement for the testflight/dev environments to use `"${admin_host}:${port}/admin/graphql"`

bodymindarts: No if statement! IMO it shouldn't be dynamic. Inject the `host` and `admin_host` in all environments. Perhaps in some (like dev) they are the same.

openoms: done in [1e677cf](https://github.com/GaloyMoney/charts/pull/2603/commits/1e677cf6c1263176a220be2a775d55829b784b30)

ntheile (approved):

openoms:
retested locally after the variable rename, merging

PR #2691, opened 2023-03-07
Kafka connect

bodymindarts:


bodymindarts:
Yes but we should remove galoy-infra

PR #2742, opened 2023-03-20
chore: use cluster-ip kafka listener instead of internal

krtk6160:


sandipndev (approved):
`cluster-ip` lol
the chart should support ClusterIP imo

krtk6160:
> `cluster-ip` lol the chart should support ClusterIP imo

The operator chart uses kebabcase for their naming convention

PR #2799, opened 2023-03-29
feat: bria chart

sandipndev:
Get https://github.com/GaloyMoney/bria into production.

### DIFF
--- a/charts/bria/templates/bria-cm.yaml
+++ b/charts/bria/templates/bria-cm.yaml
@@ -4,4 +4,17 @@ metadata:
   name: {{ include "bria.fullname" . }}-config
 data:
   bria.yml: |-
-    {{ .Values.bria.config | nindent 4 }}
+    migrate_on_start: true
+    blockchain:
+      network: {{ .Values.bria.blockchain.network }}
+    wallets:
+      sync_all_wallets_delay: {{ .Values.bria.wallets.syncAllWalletsDelay }}
+      process_all_batch_groups_delay: {{ .Values.bria.wallets.processAllBatchGroupsDelay }}
+    tracing:
+      host: {{ .Values.bria.tracing.jaegarHost }}
+      port: {{ .Values.bria.tracing.port }}
+      service_name: {{ .Values.bria.tracing.serviceName }}
+    admin:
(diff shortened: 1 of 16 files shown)


charts/bria/templates/bria-cm.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "bria.fullname" . }}-config
+data:
+  bria.yml: |-
```
bodymindarts, line 6: this won't work. Need a translation from `camelCase` <- idiomatic helm to `snake_case` <- expected in bria.
See how this is done in stablesats.

charts/bria/templates/bria-deployment.yaml
```diff
@@ -0,0 +1,63 @@
...
+        - bria
+        - daemon
+        volumeMounts:
+        - name: "config"
+          mountPath: "/bria.yml"
+          subPath: "bria.yml"
+        env:
+        - name: BRIA_CONFIG
+          value: "/bria.yml"
+        - name: BITCOIND_HOST
```
bodymindarts, line 44: These `XXX_HOST` values should be injected into config - reserve env vars for secrets.

bodymindarts (changes requested):

charts/bria/templates/bria-deployment.yaml
sandipndev: this would require source code changes: config needs to be `EnvOverride`den.

sandipndev: done

charts/bria/templates/bria-cm.yaml
```diff
@@ -0,0 +1,21 @@
...
+    migrate_on_start: true
+    blockchain:
+      network: {{ .Values.bria.blockchain.network }}
+      electrum_url: {{ .Values.bria.blockchain.electrumUrl }}
+    wallets:
```
bodymindarts, line 11: Don't add these - lets use defaults everywhere until we need to change them.

bodymindarts (approved):

PR #2836, opened 2023-04-04
ci: bump fulcrum testflight timeout

krtk6160:


krtk6160:
Closing as we're going to be using regtest bitcoind for fulcrum testflight instead of signet bitcoind.

PR #2880, opened 2023-04-13
feat: add static ip for bria api

sandipndev:
for GKE usage from bastion.

```yaml
  api:
    service:
      type: LoadBalancer
      staticIp: 10.1.2.1
      annotations:
         networking.gke.io/load-balancer-type: "Internal"
```

### DIFF
--- a/charts/bria/values.yaml
+++ b/charts/bria/values.yaml
@@ -23,7 +23,7 @@ bria:
       type: ClusterIP
       port: 2742
       staticIp: ""
-      annotations:
+      annotations: {}
 
   image:
     repository: us.gcr.io/galoy-org/bria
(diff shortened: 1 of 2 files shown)


charts/bria/values.yaml
```diff
@@ -22,6 +22,8 @@ bria:
     service:
       type: ClusterIP
       port: 2742
+      staticIp: ""
```
krtk6160, line 25:
```suggestion
      staticIP: ""
```
We use uppercase `P` in other charts with this config

krtk6160 (commented):
Couple of nits, otherwise LGTM

krtk6160 (approved):

PR #2909, opened 2023-04-17
fix: 502 intermittent errors

ntheile:
I think this should fix the 502 intermittent errors. I tested it in my local dev chart environment by force deleting the galoy api pod and running a curl on a loop (test script below). Before the nginx retry settings were applied I would get a lot of 502/503 errors when one pod was down and one pod was up. After the retry settings were applied I did not get any errors.

Screen Shot 2023-04-17 at 6 30 29 PM

ping-502.sh 
```sh
#!/bin/bash
for i in {1..5000}
do
    response=$(curl -s -o /dev/null -w "%{http_code}" http://localhost:4002/graphql)
    if [ $response -eq 502 ]
    then
        echo "502 error detected on request $i"
    else
        echo "Request $i successful, response code: $response"
    fi
    sleep 0.4
done
```

### DIFF
--- a/charts/galoy/templates/api-ingress.yaml
+++ b/charts/galoy/templates/api-ingress.yaml
@@ -15,6 +15,9 @@ metadata:
 
     nginx.ingress.kubernetes.io/proxy-read-timeout: "3600" # 1 hour
     nginx.ingress.kubernetes.io/proxy-send-timeout: "3600" # 1 hour
+    nginx.ingress.kubernetes.io/proxy-connect-timeout: "5s"
+    nginx.ingress.kubernetes.io/proxy-next-upstream: "error timeout"
+    nginx.ingress.kubernetes.io/proxy-next-upstream-tries: "3"
 
     nginx.ingress.kubernetes.io/limit-rpm: "60"
     nginx.ingress.kubernetes.io/limit-burst-multiplier: "40"
--- a/charts/galoy/templates/api-ingress.yaml
+++ b/charts/galoy/templates/api-ingress.yaml
@@ -15,7 +15,7 @@ metadata:
 
     nginx.ingress.kubernetes.io/proxy-read-timeout: "3600" # 1 hour
     nginx.ingress.kubernetes.io/proxy-send-timeout: "3600" # 1 hour
-    nginx.ingress.kubernetes.io/proxy-connect-timeout: "5s"
+    nginx.ingress.kubernetes.io/proxy-connect-timeout: "1s"
     nginx.ingress.kubernetes.io/proxy-next-upstream: "error timeout"
     nginx.ingress.kubernetes.io/proxy-next-upstream-tries: "3"
 


charts/galoy/templates/api-ingress.yaml
```diff
@@ -15,6 +15,9 @@ metadata:
 
     nginx.ingress.kubernetes.io/proxy-read-timeout: "3600" # 1 hour
     nginx.ingress.kubernetes.io/proxy-send-timeout: "3600" # 1 hour
+    nginx.ingress.kubernetes.io/proxy-connect-timeout: "5s"
+    nginx.ingress.kubernetes.io/proxy-next-upstream: "error timeout"
+    nginx.ingress.kubernetes.io/proxy-next-upstream-tries: "3"
```
nicolasburtey, line 20: by default there is no retry?

charts/galoy/templates/api-ingress.yaml
```diff
@@ -15,6 +15,9 @@ metadata:
 
     nginx.ingress.kubernetes.io/proxy-read-timeout: "3600" # 1 hour
     nginx.ingress.kubernetes.io/proxy-send-timeout: "3600" # 1 hour
+    nginx.ingress.kubernetes.io/proxy-connect-timeout: "5s"
```
bodymindarts, line 18: 5 seconds feels very long here - I would think that it should typically respond within < 1000ms. They are on the same local network.

Perhaps check honeycomb for 99 percentile latency for this hop and choose 1.5x to 2x that.

charts/galoy/templates/api-ingress.yaml
ntheile: i think it only does if you use `proxy_pass`
Screen Shot 2023-04-18 at 2 32 09 PM

openoms:
Tested this with the dev setup on the latest main and getting 502 errors even without restarting any api pods after ~60 seconds of running the ping-502 script.
 
image

No such issues when applying the changes in this PR.
Can make the requests stall by deleting one api pod, but it doesn't return 502 even after the short wait for the retry.

Can only achieve 502 if deleting both api pods at the same time.

Will try to reduce the `proxy-connect-timeout:` to 1s.

charts/galoy/templates/api-ingress.yaml
openoms: The effect of the changes are same in my local dev environment with:
```suggestion
    nginx.ingress.kubernetes.io/proxy-connect-timeout: "1s"
```

ntheile: The P99 latency is 1400ms in honeycomb. i just checked the home page, not sure if this is the right spot to check it. so `2s`  might be a good value to test @openoms

ntheile: i think the .1 second 502 makes sense, its probably the nginx rate limiting settings kicking in

openoms (approved):

bodymindarts (approved):

openoms:
Was trying to demonstrate the issue on staging, but pinging `https://api.staging.galoy.io/graphql` does not return an 502 error at all only when both api pods are deleted at the same time.

@ntheile I couldn't trigger 502 errors by deleting only one api pod neither locally or on staging.

It will be hard to know the difference on staging.

ntheile:
> Was trying to demonstrate the issue on staging, but pinging `https://api.staging.galoy.io/graphql` does not return an 502 error at all only when both api pods are deleted at the same time.
> 
> @ntheile I couldn't trigger 502 errors by deleting only one api pod neither locally or on staging.
> 
> It will be hard to know the difference on staging.

yeah makes sense. it's hard to troubleshoot transient errors. i think it might be best to just get a baseline of 502s from honeycomb over the last week, then compare it to next after after we push the fix to see if it helped

PR #3000, opened 2023-05-01
chore: add bitcoind-onchain to testflight

openoms:


ci/testflight/.gitignore
krtk6160, line 1: Not sure if this is needed here, we don't usually run tf locally in this directory

dev/bitcoin/main.tf
```diff
@@ -39,6 +39,9 @@ resource "kubernetes_secret" "bitcoind_smoketest" {
     bitcoind_endpoint    = "bitcoind.${local.bitcoin_namespace}.svc.cluster.local"
     bitcoind_port        = 18443
     bitcoind_user        = "rpcuser"
+
+    bitcoind_onchain_rpcpassword = local.bitcoind_rpcpassword
```
krtk6160, line 43: Should this be `local.bitcoind_onchain_rpcpassword` or something?

krtk6160 (changes requested):
Couple of comments

krtk6160 (approved):

ci/testflight/.gitignore
openoms: used it to validate locally

dev/bitcoin/main.tf
openoms: reusing the same password

PR #3054, opened 2023-05-09
feat: import descriptor when the secret is given

openoms:
Adding the option as empty by default:
```
descriptor:
  secretName: ""
  secretKey: ""
```

When the secret is give running the container `descriptor-import`:
```
$ k -n galoy-dev-bitcoin logs bitcoind-onchain-0 -c descriptor-import -f
# Waiting for bitcoind to finish syncing
error: timeout on transient error: Could not connect to the server 127.0.0.1:18443

Make sure the bitcoind server is running and that you are connecting to the correct RPC port.
  "initialblockdownload": false,
# Checking for the default wallet
error code: -18
error message:
Wallet file verification failed. Failed to load database path '/data/.bitcoin/regtest/wallets/default'. Path does not exist.
# Creating the default wallet
{
  "name": "default",
  "warning": ""
}
# Checking for the bitcoind-signer-descriptor
# Importing the bitcoind-signer-descriptor
[
  {
    "success": true,
    "warnings": [
      "Range not given, using default keypool range"
    ]
  },
  {
    "success": true,
    "warnings": [
      "Range not given, using default keypool range"
    ]
  }
]
# The bitcoind-signer-descriptor is present in the default wallet
```

Output on the second run when the pod is deleted:
```
$ k -n galoy-sig-bitcoin logs bitcoind-onchain-0 -c descriptor-import -f
# Waiting for bitcoind to finish syncing
error: timeout on transient error: Could not connect to the server 127.0.0.1:38332

Make sure the bitcoind server is running and that you are connecting to the correct RPC port.
  "initialblockdownload": false,
# Checking for the default wallet
{
  "name": "default",
  "warning": ""
}
# Checking for the bitcoind-signer-descriptor
  "ismine": true,
# The bitcoind-signer-descriptor is present in the default wallet
```


krtk6160 (approved):

charts/bitcoind/templates/statefulset.yaml
```diff
@@ -111,6 +111,46 @@ spec:
...
+          command: ['/bin/sh']
+          args:
+          - '-c'
+          - | 
+            echo "# Waiting for bitcoind to finish syncing"
+            while ! bitcoin-cli getblockchaininfo | grep '"initialblockdownload": false'; do
+              sleep 30
+            done
+            echo "# Checking for the default wallet"
+            while ! bitcoin-cli listwallets | grep "default"; do
```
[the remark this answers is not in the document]
openoms: was solved, only the wait for syncing is a loop now

charts/bitcoind/templates/statefulset.yaml
```diff
@@ -111,6 +111,47 @@ spec:
...
+          args:
+          - '-c'
+          - | 
+            echo "# Waiting for bitcoind to finish syncing"
+            while ! bitcoin-cli getblockchaininfo | grep '"initialblockdownload": false'; do
+              sleep 30
+            done
+            echo "# Checking for the default wallet"
+            if ! bitcoin-cli listwallets | grep "default"; then
+              if [ -d /data/.bitcoin/signet/wallets/default ]; then
```
bodymindarts, line 127: The chart knows the network right? - so we we shouldn't need this conditional.

charts/bitcoind/templates/statefulset.yaml
```diff
@@ -111,6 +111,47 @@ spec:
...
+                echo "# Loading the default wallet"
+                bitcoin-cli loadwallet "default" || exit 1
+              else
+                echo "# Creating the default wallet"
+                bitcoin-cli createwallet "default" || exit 1
+              fi
+            fi
+            echo "# Checking for the bitcoind-signer-descriptor"
+            descriptorjson=$(base64 -d /data/descriptors/descriptorjson)
+            desc=$(echo "$descriptorjson" | cut -d'"' -f 6)
```
bodymindarts, line 137: Please add comments describing input and output of these lines.

charts/bitcoind/templates/statefulset.yaml
```diff
@@ -156,6 +197,11 @@ spec:
         {{- else }}
           emptyDir: {}
         {{- end -}}
+        {{- if .Values.descriptorImport.enabled }}
```
bodymindarts, line 200: Don't think we need a flag.
Perhaps we can pass in:
`.Values.descriptor.{secretName, secretKey}` and if both are present - then that is the indication that we want to load the descriptor.

charts/bitcoind/values.yaml
```diff
@@ -139,3 +139,6 @@ bitcoindCustomConfig:
   bind: 0.0.0.0
   rpcbind: 0.0.0.0
   rpcallowip: 0.0.0.0/0
+
+descriptorImport:
+  enabled: false
```
bodymindarts, line 144: remove in favour of `descriptor.{secretName,secretKey}`

bodymindarts (approved):

charts/bitcoind/templates/statefulset.yaml
openoms: now checking if loadwallet fails proceeds to create it

charts/bitcoind/values.yaml
openoms: done, the secretName,secretKey are given as empty by default

charts/bitcoind/templates/statefulset.yaml
openoms: done, checking for a the `secretName` and `secretKey` now

charts/bitcoind/templates/statefulset.yaml
```diff
@@ -111,6 +111,47 @@ spec:
             successThreshold: 3
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+        {{- if and .Values.descriptor.secretName .Values.descriptor.secretKey }}
```
bodymindarts, line 114: Does this evaluate to true when they are set to `""`?

bodymindarts (approved):

charts/bitcoind/templates/statefulset.yaml
openoms: no, it doesn't.
Confirmed by having no `descriptor-import` container for `bitcoind-0` in dev.

krtk6160 (approved):

PR #3105, opened 2023-05-16
feat: add mempool chart and test in dev

openoms:


openoms:
Just discussed with Wiz that to use the mempool backend for fee estimation only can disable the database (done in [162798a](https://github.com/GaloyMoney/charts/pull/3105/commits/162798a74767d0b4ed956c25d2332e0c2996349d)) and connect to the bitcoind RPC only, no Electrum backend is needed: `MEMPOOL_BACKEND = "none"`

charts/mempool/templates/pvc.yaml
```diff
@@ -0,0 +1,26 @@
...
+apiVersion: v1
+metadata:
+  name: {{ template "mempool.fullname" . }}
+  annotations:
+    "helm.sh/resource-policy": keep
+  labels:
+    app: "{{ .Chart.Name }}"
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
```
krtk6160, line 12: I'm wondering if we should add these four labels to the other resources as well, or remove them from all. At the moment, it is inconsistent across our charts. For ex we use it in the [fulcrum pvc](https://github.com/galoymoney/charts/blob/ci--pin-gcp-provider-version/charts/fulcrum/templates/pvc.yaml#L9-L12) but not in its [statefulset](https://github.com/galoymoney/charts/blob/ci--pin-gcp-provider-version/charts/fulcrum/templates/statefulset.yaml#L6-L10).

I can see in the concourse chart it's being used uniformly for all resources, and we most likely got it from there.

ci/testflight/mempool/main.tf
krtk6160, line 1: Needs gcp provider version pinning like [here](https://github.com/GaloyMoney/charts/pull/3107)

ci/testflight/mempool/testflight-values.yml
```diff
@@ -0,0 +1,9 @@
+persistence:
+  enabled: false
+
+bitcoindRpcPort: 18443
+
+mariadb:
```
krtk6160, line 6: Leftover?

krtk6160 (changes requested):
Have made some comments.

Also I don't think we need the rbac stuff - role, rolebinding, svc account creation. I don't see it being used anywhere in this chart.

ci/testflight/mempool/testflight-values.yml
openoms: thanks addressed in [d049bc5](https://github.com/GaloyMoney/charts/pull/3105/commits/d049bc54c4bd46f83e99f6427d30bd76c9a4fbe5)

charts/mempool/templates/statefulset.yaml
krtk6160, line 1: I think this is missing a volumeMount for the mempool container

krtk6160 (changes requested):
One more comment

charts/mempool/templates/statefulset.yaml
openoms: this showes that it doesn't even need a volume so removed all persistence

krtk6160 (approved):
LGTM

krtk6160 (approved):

PR #3110, opened 2023-05-17
fix: remove sudo from Clean Up action

openoms:
Fixing the error:
```
Run sudo rm -rf ci/tasks/smoketest-settings
sudo: you do not exist in the passwd database
Error: Process completed with exit code 1.
```


openoms:
was included in #3109, closing.

PR #3153, opened 2023-05-22
feat: add api-deployment HPA

thevaibhav-dixit:
This PR adds Horizontal Pod Autoscaling to api-deployment.


krtk6160:
Not sure if this is needed - can reopen if it is.

PR #3163, opened 2023-05-25
chore: enable shrinkdebugfile by default

dolcalmi:
context keybase://chat/galoymoney#dev/9149


openoms:
Tested locally as well successfully. The debug log was reduced to 10 MB.

openoms (approved):

PR #3170, opened 2023-05-25
chore: add resources in kafka-cluster

thevaibhav-dixit:


charts/galoy-deps/values.yaml
```diff
@@ -9,8 +9,10 @@ strimzi-kafka-operator:
   kafka:
     listener:
       type: cluster-ip
+      resources: {}
```
krtk6160, line 12: This is scoped incorrectly, it should be under `kafka`, not `kafka.listener`.

charts/galoy-deps/templates/kafka-cluster.yaml
```diff
@@ -26,10 +26,14 @@ spec:
       log.retention.check.interval.ms: 300000 # 5 minutes
     storage:
       type: ephemeral
+    resources: 
+      {{ toYaml .Values.strimzi-kafka-operator.kafka.resources | nindent 10 }}
```
krtk6160, line 30: Indent value looks wrong

charts/galoy-deps/templates/kafka-cluster.yaml
```diff
@@ -26,10 +26,14 @@ spec:
...
   zookeeper:
     replicas: 3
     storage:
       type: ephemeral
+    resources: 
+      {{ toYaml .Values.strimzi-kafka-operator.zookeeper.resources | nindent 10 }}
```
krtk6160, line 36: Same as above, indent value is wrong

krtk6160 (changes requested):
Couple of issues

charts/galoy-deps/values.yaml
```diff
@@ -9,8 +9,10 @@ strimzi-kafka-operator:
   kafka:
     listener:
       type: cluster-ip
+    resources: {}
+  zookeeper:
+    resources: {}
   resources: {}
-
```
krtk6160, line 13: We should keep this emptyline for separation between deps

krtk6160 (approved):
Small change required otherwise LGTM

PR #3171, opened 2023-05-25
fix: always use bitcoind configmap checksum

openoms:
The predefined annotations including the configmap checksum were not applied if `.Values.podAnnotations` was empty.


krtk6160 (approved):
Ah good catch!

PR #3193, opened 2023-05-31
chore: cleanup network policy labels

krtk6160:


bodymindarts (approved):

nicolasburtey:
@krtk6160 are we no longer using a firewall? or do we have another way to do it?

krtk6160:
> @krtk6160 are we no longer using a firewall? or do we have another way to do it?

The bitcoind network policy was removed in https://github.com/GaloyMoney/charts/commit/023cbc372c4aca18ae8ffbceb52db4dc1431e06c. Discussed with @bodymindarts on removing it as it was getting in the way and it doesn't seem to offer any meaningful security benefits.

PR #3249, opened 2023-06-13
feat: device account config v3

ntheile:


ntheile:
config for https://github.com/GaloyMoney/galoy/pull/2747

charts/galoy/values.yaml
```diff
@@ -668,6 +670,31 @@ oathkeeper:
           "authorizer": { "handler": "allow" },
           "mutators": [{ "handler": "noop" }]
         },
+        {
+          "id": "device-login-routes",
+          "upstream": {
+            "url": "http://api:4002"
+          },
+          "match": {
+            "url": "<(http|https)>://<[a-zA-Z0-9-.:]+>/auth/<(create/device-account|login/device-account)>",
```
nicolasburtey, line 679: need to be narrow down to the proper link

nicolasburtey (commented):
expect https://github.com/GaloyMoney/charts/pull/3249/files#r1228648555 LGTM

UncleSamtoshi (approved):

charts/galoy/values.yaml
```diff
@@ -830,6 +849,8 @@ kratos:
   secret:
     enabled: true
     nameOverride: ""
+  image:
+    tag: v0.13.0
```
ntheile, line 853: @krtk6160 

i had to add this kratos image tag back in. In my local dev charts environment the kratos container would not boot up properly and was trying to use by default an older version of kratos 0.11.0 that does not work with out current backend codebase. any ideas?

krtk6160: @ntheile It's possible you have old charts in your `charts/charts/galoy/charts` directory. Delete all charts there, then run `helm dep up` in the `charts/charts/galoy` directory and retry.

ntheile: it worked thanks!

PR #3481, opened 2023-06-20
build: add helm-dep-updates to Makefile

openoms:
add the command in `./dev` :
```
make helm-dep-updates
```
to update all charts with dependencies.

The terraform `dependency_update  = true` is only meant for local cache and has no effect on the persisted helm chart files.

### DIFF
--- a/dev/Makefile
+++ b/dev/Makefile
@@ -127,3 +127,7 @@ review:
 delete:
 	kubectl delete namespace $(ns) --force
 
+helm-dep-updates:
+	for dir in $$(ls ../charts); do \
+		cd ../charts/$$dir && helm dependency update && cd -; \
+	done


ntheile (approved):

dev/Makefile
```diff
@@ -127,3 +127,15 @@ review:
...
+	cd bria && helm dependency update && cd .. && \
+	cd dealer && helm dependency update && cd .. && \
+	cd galoy && helm dependency update && cd .. && \
+	cd galoy-deps && helm dependency update && cd .. && \
+	cd galoy-pay && helm dependency update && cd .. && \
+	cd lnd && helm dependency update && cd .. && \
+	cd loop && helm dependency update && cd .. && \
+	cd monitoring && helm dependency update && cd .. && \
+	cd rtl && helm dependency update && cd .. && \
+	cd stablesats && helm dependency update
```
krtk6160, line 141:
```suggestion
    for dir in $$(ls ../charts); do \
		cd ../charts/$$dir && helm dependency update && cd -; \
	done
```

Couldn't we just iterate over all dirs in `/charts`?

openoms: much better and more future proof! (pushing the change because the Makefile needs tabs)

krtk6160 (changes requested):
Don't think this qualifies as a `feat`. Should probably be `build` or something.

krtk6160 (approved):

PR #3655, opened 2023-06-23
revert: build: add concurrency to dev-smoketest action (#3534)

openoms:
This reverts commit 4aa66d6f965faedd38b6e4d8907f5f2db79d7878.

The permission errors on the self hosted gh runner could be caused by one of the frequent job cancellations as a result of the concurrency setting.
Reverting to rule out as the source of the problem.
 
The self-hosted-gh-actions-runner would still likely need to be redeployed.

### DIFF
--- a/.github/workflows/dev-smoketest.yaml
+++ b/.github/workflows/dev-smoketest.yaml
@@ -3,10 +3,6 @@ on:
   pull_request:
     branches: [main]
 
-concurrency: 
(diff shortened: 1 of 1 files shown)


krtk6160:
@openoms Can we close this?

openoms:
yes, concurrency is functional now, don't need to revert any more

PR #3737, opened 2023-06-25
revert: otlp upgrades

krtk6160:


.github/workflows/dev-smoketest.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     name: Hit smoketests against Charts in Dev Setup
     runs-on: self-hosted
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v2
```
bodymindarts, line 15: didy ou mean to bundle these changes?

krtk6160: Nope, removed

bodymindarts (approved):

PR #3752, opened 2023-06-27
ci(testflight): kafka-connect to existing cluster

openoms:


charts/kafka-connect/values.yaml
```diff
@@ -1,13 +1,13 @@
 ## The image is built with the Dockerfile in /images/kafka-connect
 image:
   repository: us.gcr.io/galoy-org/kafka-connect
-  digest: "sha256:536348d31fdc138012a93b5ad54761c45075952e7c8d31508e629215e000de18"
+  digest: sha256:536348d31fdc138012a93b5ad54761c45075952e7c8d31508e629215e000de18
 
-## The pods from this namespace are allowed to access the Kafka Connect API
-smoketestNamespace: "galoy-dev-smoketest"
+# Change to allow multiple kafka-connect instances
+metadataName: kafka
```
krtk6160, line 7: What does metadata name mean? Shouldn't it be 'kafkaClusterName' or something?

charts/kafka-connect/templates/kafka-connect.yaml
```diff
@@ -1,18 +1,14 @@
 apiVersion: kafka.strimzi.io/v1beta2
 kind: KafkaConnect
 metadata:
-  name: kafka
+  name: {{ .Values.metadataName }}
```
openoms, line 4: `metadataName` is the name of the main value the variable sets. Open for suggestions for a better name,

charts/kafka-connect/values.yaml
openoms: image

How about `resourceName` ? Or any better?

charts/kafka-connect/templates/network-policy-allow-from-smoketest-namespace.yaml
```diff
@@ -2,14 +2,14 @@
...
 kind: NetworkPolicy
 metadata:
-  name: kafka-connect-allow-from-smoketest-namespace
+  name: "{{ .Values.metadataName }}-connect-allow-from-smoketest-namespace"
   namespace: {{ .Release.Namespace }}
 spec:
   podSelector:
     matchLabels:
-      strimzi.io/cluster: kafka
+      strimzi.io/cluster: {{ .Values.metadataName }}
```
krtk6160, line 10: Where are these labels set?

charts/kafka-connect/values.yaml
openoms: ok changed to `kafkaConnectInstanceName`

charts/kafka-connect/templates/network-policy-allow-from-smoketest-namespace.yaml
openoms: removed that label

charts/kafka-connect/values.yaml
openoms: changed also the `smoketstNamespace` to `allowedNamespace`

charts/kafka-connect/templates/kafka-connect.yaml
```diff
@@ -1,18 +1,14 @@
...
   version: 3.4.0
   image: "{{ .Values.image.repository }}@{{ .Values.image.digest }}"
   replicas: 1
-  {{- if .Values.bootstrapNamespace }}
-  bootstrapServers: "kafka-kafka-plain-bootstrap.{{ .Values.bootstrapNamespace }}.svc.cluster.local:9092"
-  {{- else }}
   bootstrapServers: "kafka-kafka-plain-bootstrap:9092"
-  {{- end }}
   config:
     group.id: connect-cluster
```
krtk6160, line 13: Should this be templated to avoid it being combined with the staging kafka connect?

openoms: this is an internal address within the kafka-cluster. It is the regardless of which namespace it is in. The kafka-cluster and kafka-connect needs to be in the same namespace.

openoms: I don't think that it matters as the testflight kafka-connect is not reading or writing anything so it does not interfere with the staging cluster.

PR #3757, opened 2023-06-28
ci(dev): fix dev smoketest permission issue

krtk6160:
The `dev-smoketest-settings.sh` script creates files as root and therefore the github actions, that run as a service account, do not have permissions to write over those files.


openoms (commented):
The permissions should be changed after the files are created to include them too.

openoms (approved):

PR #3760, opened 2023-06-28
ci(testflight): avoid kafka-connect name reuse

openoms:


ci/testflight/kafka-connect/main.tf
```diff
@@ -27,8 +27,8 @@ resource "kubernetes_secret" "smoketest" {
   }
 }
 
-resource "helm_release" "kafka_connect" {
-  name       = "kafka-connect"
+resource "helm_release" "kafka_connect_testflight" {
```
krtk6160, line 30: This should remain kafka_connect imo, the testflight is implicit here

krtk6160 (approved):
one nit otherwise LGTM

PR #3779, opened 2023-06-30
feat: galoy pay config for dedicated websocket server

ntheile:
to be merged with this charts config https://github.com/GaloyMoney/galoy-pay/pull/521

### DIFF
--- a/charts/galoy-pay/values.yaml
+++ b/charts/galoy-pay/values.yaml
@@ -9,7 +9,7 @@ service:
   type: ClusterIP
 graphqlHostname: domain.example
 graphqlHostnameInternal: service-name.namespace.svc.cluster.local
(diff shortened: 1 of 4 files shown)


charts/galoy-pay/values.yaml
```diff
@@ -9,6 +9,7 @@ service:
   type: ClusterIP
 graphqlHostname: domain.example
 graphqlHostnameInternal: service-name.namespace.svc.cluster.local
+websocketUrl: wss://ws.domain.example/graphql
```
krtk6160, line 12: Should we maybe call this graphqlWebsocketUrl? Seems a bit more precise and inline with the other vars

bodymindarts (approved):

PR #3828, opened 2023-07-06
feat: email login config

ntheile:
charts config for https://github.com/GaloyMoney/galoy/pull/2713

### DIFF
--- a/charts/galoy/values.yaml
+++ b/charts/galoy/values.yaml
@@ -732,7 +732,7 @@ oathkeeper:
           "id": "anonymous-rest-auth",
           "upstream": { "url": "http://api:4002" },
           "match": {
-            "url": "<(http|https)>://<[a-zA-Z0-9-.:]+>/auth/<(clearCookies|login|logout|email/code-request|email/login)>",
+            "url": "<(http|https)>://<[a-zA-Z0-9-.:]+>/auth/<(clearCookies|login|logout|email/code|email/login)>",
             "methods": ["GET", "POST", "OPTIONS"]
           },
           "authenticators": [{ "handler": "anonymous" }],
--- a/charts/galoy/values.yaml
+++ b/charts/galoy/values.yaml
@@ -729,10 +729,10 @@ oathkeeper:
     accessRules: |
       [
         {
-          "id": "cookie-anonymous-routes",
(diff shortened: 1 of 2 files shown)


charts/galoy/values.yaml
```diff
@@ -1025,6 +963,8 @@ kratos:
           verification:
             use: code # Defines which method is used, one of 'code' or 'link'.
             enabled: true
+            lifespan: 15min
```
nicolasburtey, line 966:
```suggestion
            lifespan: 15m
```

nicolasburtey:
> * figure out how to get postgres DSN for this env var `KRATOS_POSTGRES_DNS` https://github.com/GaloyMoney/galoy/pull/2713/files#diff-d33e979799a45c7c51752e9c8d96a3e452015d1a40b1e4b6ec6a98e92c4d8430R106

it's only needed for 2fa, which will really be enabled later, so we would start without it

charts/galoy/values.yaml
```diff
@@ -1035,20 +975,39 @@ kratos:
               default_browser_return_url: http://api:4002/login
       courier:
         smtp:
-          connection_uri: smtps://test:test@mailslurper:1025/?skip_ssl_verify=true
+          connection_uri: smtps://test:test@mailslurper:1025/?skip_ssl_verify=true # TODO: need to make this a secret and connect to an SMTP server we verified our domain with (sendgrid, mailgun etc...) see https://www.ory.sh/docs/kratos/emails-sms/sending-emails-smtp
+          from_address: noreply@blink.sv
+          from_name: Blink Verification Code
```
ntheile, line 980: @krtk6160 ,

can you make the `connection_uri` read from a secret? i am not sure how to get this value in a secret store (also i will be out on vacation next week). 

Thanks!

krtk6160: It looks like if we manage the kratos secrets on our own by setting the `secret.enabled=false`, which we do in deployments, it doesn't matter what we put here - just that we put something here. The smtp uri would be available to the container thru an env var which gets the value from a [secret](https://github.com/ory/k8s/blob/master/helm/charts/kratos/templates/deployment-kratos.yaml#L174). It's a bit weird the way kratos chart is written to handle this use-case.

PR #4303, opened 2023-09-19
chore: bump api ingress rate limits

krtk6160:
@nicolasburtey Not sure if the burst multiplier should be left as is

### DIFF
--- a/charts/galoy/templates/api-ingress.yaml
+++ b/charts/galoy/templates/api-ingress.yaml
@@ -15,9 +15,9 @@ metadata:
     nginx.ingress.kubernetes.io/proxy-connect-timeout: "1s"
     nginx.ingress.kubernetes.io/proxy-next-upstream: "error timeout"
     nginx.ingress.kubernetes.io/proxy-next-upstream-tries: "3"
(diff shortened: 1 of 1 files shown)


charts/galoy/templates/api-ingress.yaml
```diff
@@ -15,9 +15,9 @@ metadata:
     nginx.ingress.kubernetes.io/proxy-connect-timeout: "1s"
     nginx.ingress.kubernetes.io/proxy-next-upstream: "error timeout"
     nginx.ingress.kubernetes.io/proxy-next-upstream-tries: "3"
-    nginx.ingress.kubernetes.io/limit-rpm: "60"
-    nginx.ingress.kubernetes.io/limit-burst-multiplier: "40"
-    nginx.ingress.kubernetes.io/limit-connections: "80"
+    nginx.ingress.kubernetes.io/limit-rpm: "120"
```
nicolasburtey, line 18: this is not the settings we should change. it should be in galoy.yaml
I don't think we're really hitting the limits here, or barely (only when someone is misusing the API)

PR #4397, opened 2023-10-05
fix: admin oathkeeper request fwd

krtk6160:
This is a hacky fix for forwarding the requests made to the admin ingress to the correct path on the decisions api endpoint. Because oathkeeper uses cues from the nginx headers that we set via the auth-snippet annotation, it sends the requests to the /graphql path on the decisions api - which then gets matched against the public api rules.

### DIFF
--- a/charts/galoy/templates/admin-ingress.yaml
+++ b/charts/galoy/templates/admin-ingress.yaml
@@ -13,15 +13,15 @@ metadata:
     nginx.ingress.kubernetes.io/limit-rpm: "10"
     nginx.ingress.kubernetes.io/limit-burst-multiplier: "20"
     nginx.ingress.kubernetes.io/limit-connections: "4"
-    nginx.ingress.kubernetes.io/auth-url: "http://galoy-oathkeeper-api.{{ .Release.Namespace }}.svc.cluster.local:4456/decisions"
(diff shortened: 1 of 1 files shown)


charts/galoy/templates/admin-ingress.yaml
```diff
@@ -13,15 +13,15 @@ metadata:
     nginx.ingress.kubernetes.io/limit-rpm: "10"
     nginx.ingress.kubernetes.io/limit-burst-multiplier: "20"
     nginx.ingress.kubernetes.io/limit-connections: "4"
-    nginx.ingress.kubernetes.io/auth-url: "http://galoy-oathkeeper-api.{{ .Release.Namespace }}.svc.cluster.local:4456/decisions"
+    nginx.ingress.kubernetes.io/auth-url: "http://galoy-oathkeeper-api.{{ .Release.Namespace }}.svc.cluster.local:4456/decisions/admin"
```
nicolasburtey, line 16:
```suggestion
    nginx.ingress.kubernetes.io/auth-url: "http://galoy-oathkeeper-api.{{ .Release.Namespace }}.svc.cluster.local:4456/decisions"
```

nicolasburtey: I don't think that is needed. the request url will come from the header

nicolasburtey (approved):

charts/galoy/templates/admin-ingress.yaml
krtk6160: Yep, you're right. Just tested it and it worked without it.

PR #4400, opened 2023-10-05
chore: set hydra token ttl to 1 week

krtk6160:


nicolasburtey:
I think default settings for now, but we may want to change the setting per application later one. closing this PR for now

PR #4677, opened 2023-11-03
feat: add api keys pod

krtk6160:
- feat: add api keys pod
- chore: add apikeys image


bodymindarts (approved):

charts/galoy/values.yaml
```diff
@@ -386,6 +391,12 @@ galoy:
         failureThreshold: 5
         successThreshold: 2
         timeoutSeconds: 1
+  apiKeys:
+    nameOverride:
+    replicas: 1
```
bodymindarts, line 396: default should be 2

PR #5473, opened 2024-02-07
chore: enabling kubemonkey on all pods

nicolasburtey:


charts/galoy/values.yaml
```diff
@@ -65,7 +65,7 @@ galoy:
       repository: us.gcr.io/galoy-org/galoy-consent
       ## Digest of the image
       ##
-      digest: "sha256:be5e90bffe3ae6dc6f2729266b21022802fbed803341a90515c3ed0da6d9def1" # METADATA:: repository=https://github.com/GaloyMoney/galoy;commit_ref=84f91fb;app=consent;monorepo_subdir=apps/consent;
+      digest: "sha256:5d6ebf60a7c64aa42ccb1dd346b170be22ca5d18f02d44b249a848ba3a0ea9e4" # METADATA:: repository=https://github.com/GaloyMoney/galoy;commit_ref=d2f6f35;app=consent;monorepo_subdir=apps/consent;
```
krtk6160, line 68: What is this change?

charts/admin-panel/templates/deployment.yaml
```diff
@@ -6,6 +6,9 @@ metadata:
     app: {{ template "adminPanel.fullname" . }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     release: "{{ .Release.Name }}"
+{{- with .Values.labels }}
+{{ toYaml . | trim | nindent 6 }}
```
krtk6160, line 10: The indent looks wrong. Should be 4 I think.

charts/admin-panel/templates/deployment.yaml
```diff
@@ -17,8 +20,8 @@ spec:
       labels:
         app: {{ template "adminPanel.fullname" . }}
         release: "{{ .Release.Name }}"
-{{- with .Values.labels }}
-{{ toYaml . | trim | indent 8 }}
+{{- with .Values.podLabels }}
+{{- toYaml . | nindent 8 }}
```
krtk6160, line 24: Should be 6 here.

charts/admin-panel/values.yaml
```diff
@@ -21,3 +21,12 @@ secrets:
...
 tracingServiceName: "admin-panel"
+labels:
+  - kube-monkey/enabled: enabled
+  - kube-monkey/identifier: admin-panel
+  - kube-monkey/kill-mode: fixed
+  - kube-monkey/kill-value: "1"
+  - kube-monkey/mtbf: "3"
+podLabels:
+  - kube-monkey/enabled: enabled
+  - kube-monkey/identifier: admin-panel
```
krtk6160, line 32: Both of these should be objects, not arrays. You can validate the syntax by using `helm template admin-panel .` in `charts/charts/admin-panel`

nicolasburtey:
```
% helm template .
Error: template: admin-panel/templates/secrets.yaml:12:54: executing "admin-panel/templates/secrets.yaml" at <b64enc>: invalid value; expected string
```

krtk6160: Pushed a fix, should work now.

nicolasburtey: if I commented the value I could get it to work

charts/galoy/values.yaml
nicolasburtey: not sure why and how it showed up. reverting the change

krtk6160 (approved):

PR #5494, opened 2024-02-09
chore: adding map app

siddhart1o1:
+ [ ] make changes in deployments.


charts/map/templates/deployment.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "map.fullname" . }}
+  labels:
```
krtk6160, line 5: Need to add kubemonkey labels like here - https://github.com/GaloyMoney/charts/pull/5473

charts/map/templates/deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+spec:
+  selector:
+    matchLabels:
+      app: {{ template "map.fullname" . }}
+      release: {{ .Release.Name }}
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: {{ template "map.fullname" . }}
```
krtk6160, line 23: Need the kubemonkey labels here too

charts/map/templates/secrets.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: Secret
+metadata:
+  name: {{ template "map.fullname" . }}
+  labels:
+    app: {{ template "map.fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+type: Opaque
+data:
+  next-public-map-api-token: {{ .Values.secrets.nextPublicMapApiToken | b64enc | quote }}
```
krtk6160, line 12:
```suggestion
  map-api-key: {{ .Values.secrets.mapApiKey | b64enc | quote }}
```
We don't need the NEXT_PUBLIC prefix everywhere. It's only required by next, so it's redundant here and adds unnecessary verbosity.

krtk6160 (approved):

PR #6327, opened 2024-04-30
docs: call correct port to test the dev API

openoms:
Still results in no token:
```
{"data":{"userLogin":{"authToken":null}}}
```


openoms:
Need to test again.

PR #6631, opened 2024-06-06
chore: adding platformFeesInPpm for voucher

siddhart1o1:


ci/testflight/voucher/testflight-values.yml
```diff
@@ -6,6 +6,7 @@ voucher:
   coreUrl: "http://galoy-oathkeeper-proxy.galoy-staging-galoy.svc.cluster.local:4455/graphql"
   voucherUrl: "http://localhost:3000"
   nextAuthUrl: "http://voucher:3000"
+  platformFeesInPpm: 2000
```
krtk6160, line 9: This is redundant - it's being set in the default values

krtk6160 (approved):

PR #6975, opened 2024-08-28
chore: add getinfo check for lnd liveness probe

vindard:


krtk6160 (approved):

openoms:
Tested this locally:
```
lncli state | grep -q SERVER_ACTIVE && lncli getinfo
```
does return 0 during normal operation.

PR #7385, opened 2025-01-06
chore: Remove cala chart

k3yss:


krtk6160 (changes requested):
Also need to remove cala from the pipeline

krtk6160 (approved):

PR #7821, opened 2025-05-28
chore: changes to blink.sv domain

openoms:


openoms:
The signet.md was old, there is a more up-to-date version here: https://dev.blink.sv/self-host/deployment/signet

k9ert (approved):

PR #8094, opened 2025-08-04
feat: provisioning of IamServiceaccount credentials

k9ert:
This is:
* Provisioning a env-var called `GCP_IAM_SERVICE_ACCOUNT_PATH` with the default value of `/tmp/gcp-iam-service-account/service-account.json`
* Provisions an env-var called `GCP_PROJECT_ID` which represents the google-project, where the service-account has access

see also: https://github.com/blinkbitcoin/blink/pull/158


k9ert:
Irrelevant, closing

PR #8202, opened 2025-08-27
chore(deps): bump-lnd-backup-image-5aec38a

blinkbitcoinbot:
# Bump lnd-backup image

The lnd-backup image will be bumped to digest:
```
sha256:5d9f6da7db83976a5e8f10ad9d7cf6754dbaea1d50ec9773549e30a386eef749
```

Code diff contained in this image:

https://github.com/blinkbitcoin/charts/compare/null...5aec38a

### DIFF
--- a/charts/lnd/values.yaml
+++ b/charts/lnd/values.yaml
@@ -12,6 +12,8 @@ backupImage:
   repository: lnd-backup
   tag: latest
   pullPolicy: IfNotPresent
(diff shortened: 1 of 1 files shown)


charts/lnd/values.yaml
```diff
@@ -12,6 +12,8 @@ backupImage:
   repository: lnd-backup
   tag: latest
```
openoms, line 13:
```suggestion
```

PR #8258, opened 2025-09-29
fix: replace bitnami repo with bitnamilegacy

k9ert:
We need to switch out the bitnami charts but in the meantime, let's temporally use bitnamylegacy images.
https://github.com/bitnami/containers/issues/83267

Also:
* Adding [k9s](https://k9scli.io/) to the nix available binaries. Quite usefull for dev
* For me, Tilt went in a endless loop because after downloading the chart dependencies. The change in the Tiltfile prevents that by ignoring the directories where helm is downloading dependent charts.

### DIFF
--- a/charts/galoy/values.yaml
+++ b/charts/galoy/values.yaml
@@ -2,7 +2,9 @@
 ## This is a YAML-formatted file.
 ## Declare variables to be passed into your templates.
 
-# necessary to allow repository changes for mongodb and redis
+# necessary to allow repository changes for mongodb and redis (see below) because
+# https://github.com/bitnami/containers/issues/83267
+# ToDo: switch to non-bitnami-charts
 global:
   security:
     allowInsecureImages: true
--- a/charts/galoy/values.yaml
+++ b/charts/galoy/values.yaml
@@ -8,6 +8,13 @@ nameOverride:
 ## Provide a name to substitute for the full names of resources
 ##
 fullnameOverride:
+
+# necessary to change redis and mongodb repo sources in order to have a temp solution for:
+# https://github.com/bitnami/containers/issues/83267
+global:
+  security:
+    allowInsecureImages: true
+
 galoy:
   ## Bitcoin Network to run this Galoy instance on
   ##
@@ -599,9 +606,6 @@ galoy:
 # Therefore it doesn't have any enable/disable flag.
 #
 mongodb:
-  # necessary to allow repository changes for mongodb
-  security:
-    allowInsecureImages: true
   image:
     repository: bitnamilegacy/mongodb
   # Authentication Configuration
@@ -653,9 +657,6 @@ mongodb:
 ## Therefore it doesn't have any enable/disable flag.
 ##
 redis:
-  # necessary to allow repository changes for redis
-  security:
-  allowInsecureImages: true
   image:
     repository: bitnamilegacy/redis
   ## Redis replica config params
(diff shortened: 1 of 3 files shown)


k9ert:
tested in k3d. works!

dev/Tiltfile
dolcalmi, line 1: why is this related?

flake.nix
```diff
@@ -33,6 +33,7 @@
               yq-go
               kubernetes-helm
               opentofu
+              k9s
```
dolcalmi, line 36: same here

charts/galoy/values.yaml
dolcalmi, line 1: are we sure these are the only required changes? what about lnd? should this not update lock file?

flake.nix
k9ert: This is a tiny improvement doing as a sidetask. I guess i should have mentioned it in the description?

charts/galoy/values.yaml
k9ert: No, i'm not sure. But i think we should merge this ASAP. See the WIP issue for further details.

dev/Tiltfile
```diff
@@ -1,8 +1,10 @@
 include('./galoy-deps/Tiltfile')
 include('./bitcoin/Tiltfile')
-include('./monitoring/Tiltfile')
+#include('./monitoring/Tiltfile')
 include('./galoy/Tiltfile')
 include('./addons/Tiltfile')
-include('./stablesats/Tiltfile')
-include('./kafka-connect/Tiltfile')
+#include('./stablesats/Tiltfile')
+#include('./kafka-connect/Tiltfile')
```
openoms, line 7: uncomment these leftovers

charts/galoy/values.yaml
```diff
@@ -2,6 +2,13 @@
 ## This is a YAML-formatted file.
 ## Declare variables to be passed into your templates.
 
+# necessary to allow repository changes for mongodb and redis (see below) because
+# https://github.com/bitnami/containers/issues/83267
+# ToDo: switch to non-bitnami-charts
+global:
+  security:
+    allowInsecureImages: true
```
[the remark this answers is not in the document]
k9ert: Done

dev/Tiltfile
k9ert: Done

k9ert:
Addressed all comments but i can't test the changes right now due to:

```
 → Error: could not download oci://registry-1.docker.io/bitnamicharts/postgresql: failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/bitnamicharts/postgresql/manifests/sha256:03c06d49b0f84507c9299ad7d4eaf33ffa81c02d92d4a550bc2045d5331bca88: 429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit as 'k9ert': dckr_jti_tt7e7h1Dr5SqwqfjNjySOR4ZLuo=. You may increase the limit by upgrading. https://www.docker.com/increase-rate-limit
 ```

openoms:
the bitnamilegacy mongodb image is not available for Apple Silicon
image

k9ert:
> the bitnamilegacy mongodb image is not available for Apple Silicon
That's probably also a selling argument: If you want to develop on your mac, use the latest stuff.
If you want the full package, buy our secure images.

> Seems that the global flag is needed:
Ok, will revert.

dolcalmi (approved):

PR #8275, opened 2025-10-01
fix: use bitnamilegacy repo for postgresql images

openoms:
Follow up of: https://github.com/blinkbitcoin/charts/pull/8258
changing all bitnami images to use the bitnamilegacey repo.


k9ert (approved):
tested it and needed to fix the postgres service in galoy as well. Did it slightly different but this fix is better.

dolcalmi (approved):

PR #9061, opened 2026-05-05
fix: switch bitnami chart repo from charts.bitnami.com to OCI registry

k9ert:
## Summary
- switch Bitnami chart dependencies from `https://charts.bitnami.com/bitnami` to `oci://registry-1.docker.io/bitnamicharts`
- update matching `Chart.lock` repository entries without regenerating digests

## Notes
- `helm_release` has `dependency_update = true`, so locks can regenerate during deploy resolution
- image repository overrides are unchanged

### DIFF
--- a/.github/workflows/helm-lint.yml
+++ b/.github/workflows/helm-lint.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Get changed charts
         id: changed-charts
         run: |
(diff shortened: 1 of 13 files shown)


openoms:
The local setup is working now with updated digests.

openoms (approved):

PR #9126, opened 2026-05-19
bitcoind: retry and serialize wallet sidecar loading

### LINKED ISSUE #9125: bitcoind: make wallet-load sidecars retry instead of exiting
Wiki note: https://github.com/blinkbitcoin/blink-wip/wiki/openoms----2026-05-18-Bitcoind-Onchain-Wallet-Loading

Follow-up for the bitcoind chart wallet-loading script: retry/back off in place on transient RPC or loadwallet failures instead of exiting immediately.

openoms:
Make the bitcoind descriptor sidecars retry wallet loading in place instead of exiting on transient RPC/loadwallet failures.

Serialize wallet load/create attempts with a shared pod-local emptyDir lock so the default and withdrawal sidecars do not ask bitcoind to open wallets at the same time. Avoid createwallet attempts when the wallet path already exists.

This reduces sidecar CrashLoopBackOff risk during slow wallet recovery while keeping descriptor import failures terminal.


k9ert (approved):
LGTM Sorry for the long wait

PR #9261, opened 2026-07-08
fix(bitcoind): make liveness probe optional and shutdown grace configurable for IBD

openoms:
During initial block download bitcoind holds the RPC lock through long validation stretches, so the liveness exec probe times out and kubelet repeatedly kills a healthy, progressing node. Each kill also SIGKILLed the shutdown UTXO flush after the default 30s grace, forcing a 20-40 min chainstate replay on restart (measured on a signet node: 5 kills in 90 min, net sync progress near zero).

Add livenessProbe.enabled (default true) and terminationGracePeriodSeconds (default 600). Syncing deployments set enabled: false and re-enable once at the tip.


k9ert (approved):
LGTM

PR #9357, opened 2026-08-16
feat(blink-lnurl-server): wire proxycheck key and mode rate-limit env vars

littledino2112:
blink-lnurl-server 0.7.1 (the account-mode release) reads `LNURL_PROXYCHECK_API_KEY` to resolve country evidence for enhanced-mode accounts, plus two rate-limit tunables. This adds the chart plumbing for all three.

The key follows the existing `nsec`/`ca-cert` idiom: an `optional: true` secretKeyRef on the release secret, so environments without the key deploy unchanged and the server boots with country resolution disabled (its documented fallback — modes and gating still work, evidence columns stay NULL). The rate-limit envs render only when the values are set, keeping the server's own defaults (10/min/IP, 5000/day) authoritative rather than duplicating them in chart values.

The full path this enables was already proven on staging via a temporary `kubectl set env` override: key + egress + `x-real-ip` chain validated end to end (country evidence recorded correctly). The terraform side that feeds `proxycheck-api-key` into the secret is a companion blink-deployments PR.

### DIFF
--- a/charts/blink-lnurl-server/templates/deployment.yaml
+++ b/charts/blink-lnurl-server/templates/deployment.yaml
@@ -141,5 +141,19 @@ spec:
               name: {{ template "blink-lnurl-server.fullname" . }}
               key: "ca-cert"
               optional: true
+        - name: LNURL_PROXYCHECK_API_KEY
+          valueFrom:
+            secretKeyRef:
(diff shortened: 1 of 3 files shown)


blink-admin (approved):
Forcing this through to wire new env variables for geolocation feature

PR #9440, opened 2026-09-17
ci: enable eSign chart promotion to deployments

openoms:
The eSign chart bump job remained disabled after the downstream vendir target landed in blinkbitcoin/blink-deployments#10299. Enable it by default so successful `esign-testflight` builds can update eSign in deployments.

Update the existing render tests to require the enabled default, retain the explicit disable override, and verify that successful testflight remains mandatory. Remove the obsolete bootstrap comment.

Validation: both pipeline rendering tests and `git diff --check` pass. Verified the eSign vendir target on deployments main. Self-review found no actionable defects or additional SOLID-specific issues. Pipeline installation and live execution are not part of this PR.

Closes blinkbitcoin/blink-wip#1317


blink-admin (approved):
small ci only change

PR #9449, opened 2026-09-19
feat(galoy): expose admin api frozen mutations

littledino2112:
## What

Adds `galoy.admin.frozenMutations`, rendered as `ADMIN_API_FROZEN_MUTATIONS` on the `api` container (which also serves the admin API).

blinkbitcoin/blink#861 added that env var to core, but the chart had no way to set it, so deployments could not change the frozen list.

## Behaviour

| `galoy.admin.frozenMutations` | Rendered env | Result in core |
|---|---|---|
| unset (default) | not rendered | application default applies |
| `[]` | `ADMIN_API_FROZEN_MUTATIONS=""` | nothing frozen |
| `[a, b]` | `ADMIN_API_FROZEN_MUTATIONS="a,b"` | exactly those frozen |

A name that matches no admin mutation makes the admin server fail at startup (by design in #861), so a typo cannot silently leave a mutation open.

## Testing

Rendered the template block with helm for the three cases above (unset / `[]` / populated list); output matches the table. No default behaviour change: with the value unset the rendered deployment is identical to before.


blink-admin (approved):
Forcing the PR through

